PR #6, opened 2023-09-27
Create integration-tests.yml

manbearian:
Add some basic PR testing

The buld/test workflow is:
   1. pull a fixed commit from Triton
   2. pull current PR's triton-shared
   3. build Triton from the root
   4. run triton-share tests

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,7 +18,7 @@ jobs:
       with:
         repository: 'openai/triton'
         ref: '215b2e77a1d92f907bc4addfa3c3ba204028c32e' # known good commit "Add Shared Middle Layer to Triton via Plug-In"
-        path: ${{github.workspace}}/triton
+        path: triton
 
     - name: Update Triton to point at this build
       run: |
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -23,7 +23,8 @@ jobs:
     - name: Update Triton to point at this build
       run: |
         cd ${{github.workspace}}/triton/third_party/triton_shared
-        git switch ${{ github.sha }}
+        git fetch
+        git switch ${{ github.ref }}
 
      
    # - name: Configure CMake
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,59 @@
...
+  build_and_test_triton_shared:
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - name: Checkout Triton
+      uses: actions/checkout@v4
+      with:
+        repository: 'openai/triton'
+        ref: '215b2e77a1d92f907bc4addfa3c3ba204028c32e' # known good commit "Add Shared Middle Layer to Triton via Plug-In"
```
NathanielMcVicar, line 17: Probably not required right now, but I wonder if we want this in a file or something so we can script the checkout for local devs to make sure they are on the correct commit. This is the approach ONNX-MLIR uses for example https://github.com/onnx/onnx-mlir/blob/main/utils/clone-mlir.sh

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,59 @@
...
+        path: triton
+        submodules: 'recursive'
```
NathanielMcVicar, line 19: nit: This will checkout all the other submodules as well, which we don't need.

edit: especially since you are cloning triton-shared directly into it's third-party directory, so we don't need any submodules at all

NathanielMcVicar (approved):
LGTM other than the two minor notes above.

.github/workflows/integration-tests.yml
manbearian: yeah, not the best place for this; but i'm okay with this for now.
I wonder if Phil wouldn't mind us adding some labels...

.github/workflows/integration-tests.yml
manbearian: You're absolutely correct, but I'm going to leave this in place. The check-out takes all of 12s and this future proofs us in case we want other submodules.

PR #8, opened 2023-09-29
Improve Test Automation

manbearian:
- Add testing of latest Triton with latest Triton-Shared plugin on a daily basis
- Add testing of LKG Triton with latest Triton-Shared after updates to `main` 
- Allow testing workflow to easily be invoked with any combination Triton and Triton-Shared commits.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -5,8 +5,55 @@ on:
     branches: [ "main" ]
 
 jobs:
-  call-workflow:
-    uses: microsoft/triton-shared/.github/workflows/test-plugin.yml@v1
-    with:
-        triton-ref: '215b2e77a1d92f907bc4addfa3c3ba204028c32e' # known good commit "Add Shared Middle Layer to Triton via Plug-In"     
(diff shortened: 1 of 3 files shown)


NathanielMcVicar (approved):
LGTM, this is very clean! I wanted to verify the manual workflow_dispatch worked, but I couldn't figure out how from the PR. I guess if it's broken we can just have another PR to fix it.

manbearian:
> LGTM, this is very clean! I wanted to verify the manual workflow_dispatch worked, but I couldn't figure out how from the PR. I guess if it's broken we can just have another PR to fix it.

Yeah, as far as i can tell you cannot run the workflow that isn't in main yet (there's an option to run a different version from another branch, but not a new workflow)

manbearian:
@NathanielMcVicar , so far looks good:
 - New workflow worked in a PR (this one)
 - New workflow triggered after this PR merged into main
 - manual workflow works: i tested main/main 
 - we'll see if scheduled (nightly) job works, but it should since i already manually tested main/main which is what the scheduled job will invoke.

PR #9, opened 2023-10-11
Update triton-shared 10/2023

nhat-nguyen:
This update includes various fixes and the following features:
- preliminary support for block-pointers
  - block-pointers are lowered to memref buffers similarly to traditional triton pointer loads
- support lowering for `triton.get_num_programs` by introducing extra arguments when launching triton kernels similarly to `triton.get_program_id`
- improved lowering `triton.reduce` which previously only supports lowering float values
- preliminary support for pointer arithmetic involving the modulo operator
  - the modulo operator use case is seen in the tutorial matmul example, where pointer offsets are being modded to prevent loading out-of-bound values; in such case, these values are wrapped around to the beginning of the buffer

### DIFF
--- a/include/triton-shared/Analysis/MaskAnalysis.h
+++ b/include/triton-shared/Analysis/MaskAnalysis.h
@@ -14,6 +14,8 @@
 
 #include "triton/Dialect/Triton/IR/Dialect.h"
 
+#include <utility>
+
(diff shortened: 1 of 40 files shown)


manbearian (approved):
LTGTM

PR #19, opened 2023-10-16
remove unused 'triton-shared-translate'

manbearian:
We hit a build break in upstream Triton due to an API change. The build break occurred in `triton-shared-translate`. Since we don't use this binary i'm completely removing it rather than fixing the break.

## Testing
I confirmed live-live (head of Triton built with head of triton-shared) builds correctly with this change.


manbearian:
live-live build:
https://github.com/microsoft/triton-shared/actions/runs/6536803141

nhat-nguyen (approved):
Looks good to me! What is this triton-shared-translate used for by the way?

manbearian:
> Looks good to me! What is this triton-shared-translate used for by the way?

Well its not used, that's why i deleted it :)
But i think the answer you're looking for is that it ~is~ was this project's analogue to `mlir-translate` (a tool to convert MLIR to LLVMIR).

PR #36, opened 2023-11-02
Fix Several Problems with Auto-Issue Creation

manbearian:
- Add some logic to reuse existing rolling build failure bug if one exists.
- Fix misuse of invoking workflow causing rolling build failure
- several other QoL improvements, cleanup

Users are still not getting notifications :(


manbearian:
> abandoning for now... wil retry later

nhat-nguyen (approved):
Looks good! Would be great if we can share the "Report failure" step if it's supported 😄

PR #41, opened 2023-11-06
Fix nightly workflow

nhat-nguyen:
The nightly workflow is broken because GitHub doesn't populate the default values for the `inputs` variable, resulting in us passing in an empty string for a boolean input. As a workaround, I check if the user-provided input contains the `true` string instead (Github automatically converts the boolean to a string representation when using the `contains` function). Comparing with `true` directly is cumbersome according to this: https://github.com/actions/runner/issues/1483

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -25,7 +25,7 @@ jobs:
   report-failure:
     runs-on: ubuntu-latest
     needs: [call-workflow]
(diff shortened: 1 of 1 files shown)


manbearian (commented):
Seems okay. It would be nice if we could test this stuff, but i think just poking around is okay since it's not a big deal as the project is pretty small right now.

manbearian (approved):

PR #43, opened 2023-11-06
Remove the report failure step in `test-plugin.yml`

nhat-nguyen:
"Report failure" is not working in `test-plugin.yml` because we don't have `inputs.notify-on-failure` defined in the file. But we don't need this step in `test-plugin.yml` because the only two workflows that include it are `integration-tests.yml` and `nightly-tests.yml`; `integration-tests.yml` run on pull requests and merges to main, so there's no need to report failures. `nightly-tests.yml` on the other hand already has the report failure job in it.


.github/workflows/test-plugin.yml
manbearian, line 1: this code should never have been checked-in. my bad. Testing YML is hard 😁

manbearian (approved):

PR #45, opened 2023-11-08
Add `TritonTilingExt` dialect to support ops not representable in linalg

nhat-nguyen:
The `TritonTilingExt` dialect leverages linalg's TilingInterface to add tiling & fusion support for operators that cannot be represented in linalg Much of the tiling implementation is borrowed from linalg's tiling code. As an example, I have added a barebone version of `cumsum` that represents the cumulative sum of the inner-most dimension of a tensor. Other backends can then lower the op to lower-level implementation as needed after applying tiling & fusion.

### DIFF
--- a/include/triton-shared/CMakeLists.txt
+++ b/include/triton-shared/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(Conversion)
+add_subdirectory(Dialect)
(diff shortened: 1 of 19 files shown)


manbearian (approved):
Fantastic! This is a great example of how to represent operations in the middle-layer without relying directly on the linalg dialect.

manbearian:
The `TritonTilingExt` dialect leverages linalg's `TilingInterface` to add tiling & fusion support for operators that cannot be represented in linalg Much of the tiling implementation is borrowed from linalg's tiling code. As an example, I have added a barebone version of `cumsum` that represents the cumulative sum of the inner-most dimension of a tensor. Other backends can then lower the op to lower-level implementation as needed after applying tiling & fusion.

PR #51, opened 2023-11-14
Fix cpu import error

yuanfz98:
From triton commit 768fc1fcd98ecfc0892f8982b0bb009dd7bb11ea, the functon `version_key()` is no longer there.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -341,7 +341,7 @@ class TritonSharedRefCPUBackend(BaseBackend):
 
     def make_launcher_stub(self, name, signature, constants, ids):
         # name of files that are cached
-        so_cache_key = make_so_cache_key(self.version_key, signature, constants, ids)
+        so_cache_key = make_so_cache_key(self.get_version_key(), signature, constants, ids)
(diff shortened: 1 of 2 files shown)


nhat-nguyen:
Thanks @yuanfz98, would you mind also updating the commit hash in `.github/workflows/integration-tests.yml` as well? Otherwise the change looks good to me.

python/__init__.py
```diff
@@ -336,7 +341,7 @@ def get_architecture_descriptor(self, **kwargs):
 
     def make_launcher_stub(self, name, signature, constants, ids):
         # name of files that are cached
-        so_cache_key = make_so_cache_key(version_key(), signature, constants, ids)
+        so_cache_key = make_so_cache_key(self.version_key, signature, constants, ids)
```
nhat-nguyen, line 344: this should be `self.get_version_key()` because `self.version_key` is initially `None`

nhat-nguyen (changes requested):

nhat-nguyen (approved):

PR #62, opened 2023-11-22
[WIP] add a canonicalizer before triton-to-linalg

yuanfz98:
This PR is WIP and aims to add a canonicalizer for triton-to-linalg. It decouples the mutation of ttir from triton-to-linalg. 
RemsiCanonicalizer will postpone expand_dims{axis=1} and provide valid input for PtrAnalysis, as the latter checks rank == 1:
```
void PtrAnalysis::visitOperandRem(
    arith::RemSIOp remOp, PtrState &state, const Location loc,
    ConversionPatternRewriter &rewriter,
    const llvm::SmallDenseMap<Value, PtrState> &knownPtrs) {
  assert(state.isEmpty());
  visitOperand(remOp.getLhs(), state, loc, rewriter, knownPtrs);
  assert(state.getRank() == 1 && !state.modulos.back().has_value() &&
         "No support for multiple modulos within an expression");
``` 
After RemsiCanonicalizer, %11 = arith.remsi %5, %cst_11 : tensor<8x1xi32> will be :
```
%11 = arith.remsi %5_clone, %cst_clone : tensor<8xi32>
%12 = tt.expand_dims %11 {axis = 1 : i32} : (tensor<8xi32>) -> tensor<8x1xi32>
```
Thus it is no longer a rank 2 tensor for PtrAnalysis.

### DIFF
--- a/include/triton-shared/CMakeLists.txt
+++ b/include/triton-shared/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(Conversion)
+add_subdirectory(Dialect)
--- a/include/triton-shared/Dialect/CMakeLists.txt
+++ b/include/triton-shared/Dialect/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(TritonTilingExt)
--- a/include/triton-shared/Dialect/TritonTilingExt/CMakeLists.txt
+++ b/include/triton-shared/Dialect/TritonTilingExt/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(IR)
--- a/include/triton-shared/Dialect/TritonTilingExt/IR/CMakeLists.txt
+++ b/include/triton-shared/Dialect/TritonTilingExt/IR/CMakeLists.txt
@@ -1,0 +1,11 @@
+set(LLVM_TARGET_DEFINITIONS TritonTilingExtOps.td)
+mlir_tablegen(TritonTilingExtOpsDialect.h.inc -gen-dialect-decls -dialect=ttx)
+mlir_tablegen(TritonTilingExtOpsDialect.cpp.inc -gen-dialect-defs -dialect=ttx)
+mlir_tablegen(TritonTilingExtOps.h.inc -gen-op-decls)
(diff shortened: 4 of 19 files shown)


nhat-nguyen:
Thank you for the contribution. I have a small patch to support the modulo pattern that torch inductor generates together with some other small fixes. I will take a look at your proposal for canonicalizing the division operator. Thanks!

nhat-nguyen:
I really like the idea of canonicalizing before TritonToLinalg so that PtrAnalysis only needs to take care of one pattern. I think for now though, to support the case that you're interested in, the code itself is quite short so we can still have it in PtrAnalysis to keep the complexity low. I have a working branch over at `nhat/modulo` if you're interested in checking out early. I just need to do a bit of cleanup before publishing the PR. Here's the code to support your case:

```cpp
if (state.getRank() == 1) {
    // Apply the modulo pattern before expanding shape, the common pattern is
    // offs_am = (pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M)) % M
    // a_ptrs = a_ptr + (offs_am[:, None] * stride_am + offs_k[None, :] *
    // stride_ak)

    assert(!state.modulos.back().has_value() &&
           "No support for multiple modulos within an expression");

    state.modulos.back() = ModuloState{rhsState.scalar};

  } else if (state.getRank() == 2) {
    // torch inductor expands the tensor shape before applying the modulo
    // pattern.
    //
    // We only support either:
    // - (tl.arange(0, end)[:, None] % mod), or
    // - (tl.arange(0, end)[None, :] % mod)
    //
    // In both cases, we apply the modulo to the non-singleton dimension.
    auto shape = cast<TensorType>(remOp.getResult().getType()).getShape();
    if (shape[0] == 1) {
      state.modulos[1] = ModuloState{rhsState.scalar};
    } else if (shape[1] == 1) {
      state.modulos[0] = ModuloState{rhsState.scalar};
    } else {
      assert(false && "Do not support taking modulo on a 2D tensor with no "
                      "singleton dimension");
    }
  } else {
    assert(false && "Unsupported modulo pattern");
  }
```

PR #67, opened 2023-11-29
Add matmul kernel test for CPU backend

nhat-nguyen:
Add the latest matmul triton tutorial as a test for the CPU backend. This exposed a bug in the TritonToLinalg pass where we need to initialize the init tensor with all zeroes first since allocated buffers aren't guaranteed to be zero-initialized.


python/examples/test_matmul.py
manbearian, line 1: i was hoping to avoid replicating the entire tutorial code. Is there a way we can import this?

manbearian (approved):
Looks good, my only feedback is around looking for a way to avoid replicating the existing tutorial code.

python/examples/test_matmul.py
nhat-nguyen: yea this turns out harder than i expected:
1) we need to modify the device type from `cuda` to `cpu` in the original code anyway
2) our project is independent from the tutorial code which lives in the `triton` repo, so we can't directly access those files anyway

manbearian:
Yes, no need to fix here... but we really need a way to address this... I wonder if we can modify the tutorials themselves with some conditional code to check if CPU is supported and test CPU back-end if it is?

ian Bearman
Principal Software Engineering Manager
AI Frameworks: Compilation and Optimization (Maia 100<https://azure.microsoft.com/en-us/blog/microsoft-azure-delivers-purpose-built-cloud-infrastructure-in-the-era-of-ai/> and beyond!)
Now hiring!<https://aka.ms/ianb.jobs>
#BlackLivesMatter
/* Making your AI faster, cheaper, smarter! */

PR #75, opened 2023-12-26
Extend MinMaxConverter with OGE and OLE cmpf predicates

aobolensk:
Convert arith.cmpf operation with OGE and OLE predicates the same way as for OGT and OLT to max/min operation


aobolensk:
@microsoft-github-policy-service agree

nhat-nguyen (approved):
Thank you!

PR #82, opened 2024-01-10
Introduce TritonStructured dialect and triton-to-structured pass

haishanzzzz:
This PR introduces TritonStructured dialect and `triton-to-structured` pass. Please see #81 for background.

This PR is broken into 5 commits:
1. Introduce TritonStructured dialect, which includes three ops: `tts.make_tptr`, `tts.load`, and `tts.store`
2. Update the current MaskAnalysis, which removed unused functions that produces different flavors of subviews
3. Minor updates to OpFoldResultsUtils
4. Introduce `triton-to-structured` pass
5. Adds LIT tests

A few notes:
- `triton-to-structured` does not use `DialectConversion` but rather manually walks the IR to perform rewriting. The reason is this pass does not try to legalize Triton operations (in fact it allows them to live after the pass) but rather opportunistically rewrite certain ops if the analysis succeeds. Legality marking in `DialectConversion` does not provide a way to set certain instances of an op as legal, which makes it a poor fit for our purpose.
- Compared to `triton-to-linalg`, `triton-to-structured` is more robust when encountering IRs that it cannot analyze. Specifically, it `emitRemark` when analysis fails, `emitWarning` when it may (although very unlikely) produce wrong results, and only fail when a logic error is detected. This graceful failure can be demonstrated by manually rerunning `wraparound_unsupported_add_offset.mlir`.
- LIT tests are directly borrowed from the current tests for `triton-to-linalg`. Unrelated tests are removed and output is manually examined to verify correctness (except for those ones generated for Triton tutorials).
- `tt.make_tptr` will only be used to handle tensor of pointers.
- When pointer analysis fails, or when the original code contains scalar pointer accesses, `tt.addptr` will be left in the output of the pass. They should be converted to memref operations in `structured-to-memref`.
- Analysis code is unfortunately duplicated for the time being. Please bear with the awkward names.

A main follow up to this PR is to get consensus on how we'd like to handle block pointers in this pass. 

There are a few other misc. items on the wishlist for the new code:
- Cache pointer and mask analysis results to avoid introducing duplicated ops
- More testing on wrap around behavior
- Pretty printer, verifier, canonicalizer for newly introduced ops
- Investigate why `tt.reduce` make `--remove-dead-values` pass fail


include/triton-shared/Analysis/OpFoldResultUtils.h
```diff
@@ -1,12 +1,12 @@
...
 //
-// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation, Meta Platforms.
 // Licensed under the MIT license.
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef TRITON_ANALYSIS_OPFOLDRESULT_UTILS_H
-#define TRITON_ANALYSIS_OPFOLDRESULT_UTILS_H
+#ifndef TRITON_ANALYSISSTRUCTURED_OPFOLDRESULT_UTILS_H
```
haishanzzzz, line 8: TODO: revert this

nhat-nguyen:
Thank you @haishanzzzz, I'll start taking a look :)

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  return success();
+}
+
+LogicalResult PtrAnalysis::rewriteOp(Operation *rootOp) {
+  LLVM_DEBUG({
+    llvm::dbgs() << "rewriting rootOp\n";
+    rootOp->dump();
+  });
+
+  rootOp->walk<WalkOrder::PreOrder>([&](Operation *op) {
```
nhat-nguyen, line 1089: > triton-to-structured does not use DialectConversion but rather manually walks the IR to perform rewriting. The reason is this pass does not try to legalize Triton operations (in fact it allows them to live after the pass) but rather opportunistically rewrite certain ops if the analysis succeeds. Legality marking in DialectConversion does not provide a way to set certain instances of an op as legal, which makes it a poor fit for our purpose.

I believe in this case we can make use of the [Pattern Rewrite Driver](https://mlir.llvm.org/docs/PatternRewriter/#greedy-pattern-rewrite-driver) (`applyPatternsAndFoldGreedily`). The driver will repeatedly call these patterns until the patterns fail to match (i.e.: `return failure()`). I think this fits quite well with most of the rewrite functions that we have here. `rewriteForOp` might be a trickier because now we need to determine if the `scf::ForOp` has already been rewritten. Would be nice if we can use this, but otherwise if it's too much complexity, I think it's fine to use the approach we have here.

include/triton-shared/AnalysisStructured/PtrAnalysis.h
```diff
@@ -0,0 +1,209 @@
...
+namespace triton {
+
+const extern std::string ptrAnalysisAttr;
+
+// Data structure used to decode pointer arithmetics. offsets, sizes, and
+// strides are in unit of elements in a linearly laid-out memory, which is the
+// same as pointer arithmetic operations in Triton language. scalar is a
+// shortcut used when the entire state describes a single scalar value. source
+// is the base pointer.
+class PtrSState {
```
nhat-nguyen, line 33: Curious about the name, is `PtrSState` intentional or a typo? 😄

nhat-nguyen: Oh nvm, I see that we are duplicating the analysis so this is to prevent name collision

haishanzzzz:
> @haishanzzzz please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: That's a very good point. `rewriteForOp` was the thing that gave me most headache. The reason is that it can potentially produce operations that we may or may not want to rewrite. With `DialectConversion` it's not easy to say "I tried rewriting this Op but failed and I am ok with leaving it as is". It looks like `PatternRewriter` is probably a better fit. But tbh it takes so much time to try something like this I kinda don't want to spend the time.

Another potentially tricky thing is we don't always want to replace the results of `addptr`. E.g., if we have a chain of addptr->addptr->load. We can rewrite the first addptr but not the second one, we'd like to keep the first one as is. It's unclear if `PatternRewriter` can handle all this. I believe the manual walk + custom data structure provides the most flexibility. (This is currently handled by the additional IRMapping defined in PtrAnalysis struct.)

include/triton-shared/AnalysisStructured/PtrAnalysis.h
manbearian: isn't this what namespaces are for?

manbearian:
@haishanzzzz please , can you tell me more about this:
> Investigate why tt.reduce make --remove-dead-values pass fail

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+  for (uint64_t i = 0; i < lhs->getRank(); i++) {
+    if (!lhs->dimHasModulo(i)) {
+      modulos.push_back(lhs->modulos[i]);
+    } else if (hasConstZero(rhs->offsets[i])) {
+      modulos.push_back(lhs->modulos[i]);
+    } else if (i == 0 && lhs->getRank() == 2 && rhs->scalar) {
+      modulos.push_back(lhs->modulos[1]);
+      modulos.push_back(lhs->modulos[0]);
+      op->emitWarning(
```
haishanzzzz, line 200: Should we add a strictly-correct option and fail this case? Afaik this is the only transformation in this pass that can lead to potentially incorrect result.

haishanzzzz:
> @haishanzzzz please , can you tell me more about this:
> 
> > Investigate why tt.reduce make --remove-dead-values pass fail

This is technically not related to this PR, but if I just do `triton-opt --remove-dead-values` with tt.reduce things actually fail.
```
triton-opt --remove-dead-values reducemax_32_256_bf16.mlir                                                                                                                                                                    16:57:35
reducemax_32_256_bf16.mlir:35:7: error: null operand found
      tt.reduce.return %22 : bf16
      ^
reducemax_32_256_bf16.mlir:35:7: note: see current operation: "tt.reduce.return"(<<NULL VALUE>>) : (<<NULL TYPE>>) -> ()
```

Is this what you see too?

lib/AnalysisStructured/MaskAnalysis.cpp
```diff
@@ -0,0 +1,303 @@
+//===----------------------------------------------------------------------===//
+//
+// Copyright (c) Microsoft Corporation, Meta Platforms.
+// Licensed under the MIT license.
+//
+//===----------------------------------------------------------------------===//
+
```
nhat-nguyen, line 7: I just diff'd this with the original MaskAnalysis file, it looks like the two only changes are:

1. the last arg's type changed from `ConversionPatternRewriter` to `OpBuilder`
2. removed `getExtractSlice`, `getSubview`, `getSideBySideSubviews` and `getStackedSubviews`

With only these two changes, I think we still can reuse the original file similarly to `OpFoldResultUtils` without duplicating it by simply changing the last's arg type to `OpBuilder`. `OpBuilder` is a superclass of `ConversionPatternRewriter`, so the original `TritonToLinalg` pass will still work as is.

This will definitely help reduce the complexity when it's time to retire the monolith approach.

lib/AnalysisStructured/MaskAnalysis.cpp
nhat-nguyen, line 1: I just diff'd this with the original MaskAnalysis file, it looks like the two only changes are:

the last arg's type changed from ConversionPatternRewriter to OpBuilder
removed getExtractSlice, getSubview, getSideBySideSubviews and getStackedSubviews
With only these two changes, I think we still can reuse the original file similarly to OpFoldResultUtils without duplicating it by simply changing the last's arg type to OpBuilder. OpBuilder is a superclass of ConversionPatternRewriter, so the original TritonToLinalg pass will still work as is.

This will definitely help reduce the complexity when it's time to retire the monolith approach.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -0,0 +1,137 @@
...
+
+def TTS_MakeTensorPtrOp
+  : TTS_Op<"make_tptr", [ AttrSizedOperandSegments, Pure]> {
+  let summary = "create a pointer that points to a tensor in memory";
+
+  let arguments = (ins TT_Ptr:$base,
+                       DenseI64ArrayAttr:$sizes,
+                       Variadic<Index>:$strides,
+                       Variadic<Index>:$offsets,
+                       Variadic<Index>:$parent_sizes,
```
nhat-nguyen, line 41: I think the other variables are self-explanatory, but for `parent_sizes`, it might be worth mentioning that it's for tracking the sizes of the parent used in a modulo expression

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -0,0 +1,137 @@
...
+
+//
+// Op Base
+//
+class TTS_Op<string mnemonic, list<Trait> traits = []> :
+    Op<Triton_Structured_Dialect, mnemonic, traits> {
+}
+
+def TTS_MakeTensorPtrOp
+  : TTS_Op<"make_tptr", [ AttrSizedOperandSegments, Pure]> {
```
nhat-nguyen, line 34: nit: extra space before `AttrSizedOperandSegments`

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
```diff
@@ -0,0 +1,71 @@
...
+                            ArrayRef<OpFoldResult> parentSizes) {
+  SmallVector<int64_t> staticStrides, staticOffsets, staticParentSizes;
+  SmallVector<Value> dynamicStrides, dynamicOffsets, dynamicParentSizes;
+
+  dispatchIndexOpFoldResults(offsets, dynamicOffsets, staticOffsets);
+  dispatchIndexOpFoldResults(strides, dynamicStrides, staticStrides);
+  dispatchIndexOpFoldResults(parentSizes, dynamicParentSizes,
+                             staticParentSizes);
+
+  auto basePtr = base.getType().cast<triton::PointerType>();
```
nhat-nguyen, line 33: nit: mlir is looking to at one point deprecate this style of casting, prefer to use `cast<triton::PointerType>(base.getType())` instead 😄

lib/AnalysisStructured/PtrAnalysis.cpp
nhat-nguyen: This sounds good to me. Not worth it to spend time catering to the PatternRewriter.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  }
+
+  for (uint64_t i = 0; i < lhs->sizes.size(); i++) {
+    OpFoldResult newOffset =
+        mulOFRValue(lhs->offsets[i], rhs->scalar, loc, builder);
+    OpFoldResult newStride =
+        mulOFRValue(lhs->strides[i], rhs->scalar, loc, builder);
+    OpFoldResult newModulo =
+        mulOFRValue(lhs->modulos[i], rhs->scalar, loc, builder);
+    offsets.push_back(newOffset);
```
nhat-nguyen, line 257: We had this tracked in an issue already, but I think this is better if we fail the analysis if we are multiplying an expression containing modulo with a scalar. Let's take the following example:

Assume we have a tensor with `cols = 4`

The user wants to load a `1x3` tensor starting at offset 2 while also modding it with 4

```
(2 + tl.arange(0, 3)) % 4
= (2 + [0, 1, 2]) % 4
= ([2, 3, 4] % 4)
= [2, 3, 0]

   0    1    2    3
0  x         x    x
1
2
3
```

This is all good.

But now the user wants to multiply the offsets by 2:

```
((2 + tl.arange(0, 3)) % 4) * 2
= [2, 3, 0] * 4
= [8, 12, 0] <----- these offsets are completely non-sense
```

These offsets are now all over the place, and us multiplying the mod (`4`) with `2` doesn't make sense either.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        continue;
+      } else if (srcShape[i] < dstShape[i]) {
+        state.sizes[i] = builder.getIndexAttr(dstShape[i]);
+      } else {
+        llvm_unreachable("unexpected dimensions used in broadcast");
+      }
+    }
+    return success();
+  }
+
```
nhat-nguyen, line 478: nit:
- this code looks quite nested; i think we can reverse the condition `src.getType().isa<ShapedType>()` and fail early
- `srcShape` declaration and assignment should be on the same line, here it's split but the initial value isn't used elsewhere
- should probably change to `cast<ShapedType>(src.getType())` similar to my other comment since this style cast isn't preferred now

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  if (src.getType().isa<IntegerType, IndexType, triton::PointerType>()) {
+    for (auto s : dstShape) {
+      state.offsets.push_back(builder.getIndexAttr(0));
+      state.sizes.push_back(builder.getIndexAttr(s));
+      state.strides.push_back(builder.getIndexAttr(0));
+      state.modulos.push_back(builder.getIndexAttr(0));
+    }
+  } else {
+    splatOp->emitRemark("PtrAnalysis: unsupported splat pattern");
```
nhat-nguyen, line 505: The original code in the else clause is the following, which is admittedly complicated:

```cpp
    // src is a memref that represent a scalar pointer; it should have
    // one dimension of size 1. This happens inside a for loop that
    // originally has an init arg that is a tensor of pointers; this arg
    // would have been replaced by rewriteForOp.
    auto srcType = src.getType().cast<MemRefType>();
    assert(srcType.getRank() == 1 && state.getRank() == 1 &&
           "splat MemRef source should have rank 1");
    assert(srcType.getShape()[0] == 1 &&
           getIntAttr(state.sizes[0]).value() == 1 &&
           "splat MemRef source should have size 1");

    // Stride[0] will have value of 1 set in visitOperandAddPtr. This
    // value will be represented by a constOp. Clear this value.
    state.strides[0] = rewriter.getIndexAttr(0);

    for (auto [i, s] : llvm::enumerate(dstShape)) {
      if (i == 0) {
        state.sizes[i] = rewriter.getIndexAttr(s);
        continue;
      }
      state.offsets.push_back(rewriter.getIndexAttr(0));
      state.sizes.push_back(rewriter.getIndexAttr(s));
      state.strides.push_back(rewriter.getIndexAttr(0));
      state.modulos.push_back(std::nullopt);
    }
```

Is this something we won't support anymore?

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  }
+
+  PtrSState offsetState;
+  if (visitOperand(addptrOp.getOffset(), offsetState, addptrOp.getLoc(),
+                   builder)
+          .failed()) {
+    return failure();
+  }
+
+  assert(ptrState.source && "ptr field should provide source / base pointer");
```
nhat-nguyen, line 542: Similar question, I see this block gets removed too:

```cpp
  // Handle the special case when we are in a for loop, ptr is originally a
  // scalar pointer but replaced with a memref. In this case, ptrState will have
  // rank 1 and offsetState will have rank 0.
  // TODO:
  //  Passing a block argument pointer directly into a for loop not supported
  if (ptrState.getRank() == 1 && offsetState.getRank() == 0) {
    offsetState.sizes.push_back(rewriter.getIndexAttr(1));
    offsetState.offsets.push_back(offsetState.scalar);
    offsetState.strides.push_back(rewriter.getIndexAttr(0));
    offsetState.modulos.push_back(std::nullopt);
  }
```

Is this something we won't support

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  if (operand.getType().isa<triton::PointerType>()) {
+    // A scalar pointer can either be produced by AddPtrOp or a block
+    // argument
+    if (auto op = operand.getDefiningOp()) {
+      if (auto addPtrOp = dyn_cast<triton::AddPtrOp>(op)) {
+        return visitOperandAddptr(cast<triton::AddPtrOp>(op), state, loc,
+                                  builder);
+      } else if (auto makeTensorOp = dyn_cast<triton::MakeTensorPtrOp>(op)) {
+        llvm_unreachable("NYI");
+      } else {
```
nhat-nguyen, line 631: let's add a github issue to track this

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  if (visitOperandAddptr(op, state, op.getLoc(), builder).failed()) {
+    return failure();
+  }
+
+  knownPtrs[op.getResult()] = state;
+
+  if (op.getPtr().getType().isa<RankedTensorType>()) {
+    auto maketptrOp = state.createTTSMakeTensorPtrOp(builder, op.getLoc());
+    map.map(op.getResult(), maketptrOp.getResult());
+  } else {
```
nhat-nguyen, line 679: i like this, awesome :)

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    map.map(op.getResult(), op.getResult());
+  }
```
nhat-nguyen, line 681: is this map redundant since it's just mapping to itself?

nhat-nguyen: oh i see we're using this map in `rewriteForOp` to also check if a value is pointer related

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+      }
+    }
+    // Init arg is not pointer related or prior rewrite has failed. Pass as is
+    else {
+      newInitArgs.push_back(arg);
+      continue;
+    }
+    // Record the PtrSState for later processing
+    initArgIndexState.push_back(std::make_pair(i, state));
+  }
```
nhat-nguyen, line 725: should we only push if the state is not empty? the only two cases where i think this is necessary are when we call `visitOperandMakeTPtr` or `visitOperandAddptr`

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+          continue;
+        }
+        newInitArgs.push_back(mappedV);
+      }
+    }
+    // Init arg is not pointer related or prior rewrite has failed. Pass as is
+    else {
+      newInitArgs.push_back(arg);
+      continue;
+    }
```
nhat-nguyen, line 722: I think this code is a bit hard to follow even though its intention is simple.

So the conditions required to use the remapped value if it exists is when:
- the defining op of the value is either AddPtrOp or MakeTensorPtrOp, and
- we have successfully visited the op.

For all other cases, we just reuse the original arg.

I would probably simplify it to something like

```
if (mappedV) {
    if (auto makeTPtrOp = mappedV.getDefiningOp<tts::MakeTensorPtrOp>()) {
        if (visitOperandMakeTPtr(makeTPtrOp, state, op.getLoc(), builder)
                .succeeded()) {
            newInitArgs.push_back(mappedV);
          continue;
        }
      } else if (auto addptrOp = mappedV.getDefiningOp<triton::AddPtrOp>()) {
        assert(!addptrOp.getResult().getType().isa<RankedTensorType>());
        if (visitOperandAddptr(addptrOp, state, op.getLoc(), builder)
                .succeeded()) {
          newInitArgs.push_back(mappedV);
          continue;
        }
      }
}

newInitArgs.push_back(arg);

// Record the PtrSState for later processing
...
```

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+LogicalResult PtrAnalysis::rewriteForOp(scf::ForOp op) {
+  SmallVector<Value> newInitArgs;
+
+  SmallVector<std::pair<int, PtrSState>, 5> initArgIndexState;
+  SmallVector<std::pair<int, PtrSState>, 5> knownPtrsTmp;
+
+  llvm::SmallDenseMap<int, PtrSState> initArgIndexMap;
+
+  OpBuilder builder(op);
+
```
nhat-nguyen, line 694: Just a quick sanity check: does this set the insertion point right before the op

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+        state.strides[j] = constOp.getResult();
+      } else {
+        newInitArgs.push_back(s.get<Value>());
+      }
+    }
+
+    if (state.getRank() == 0) {
+      assert(state.scalar);
+      newInitArgs.push_back(state.scalar);
+    }
```
nhat-nguyen, line 761: this code is new, can we have an example snippet describing the case? i'm not quite following why we have to push the scalar to new init args

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+      auto addptrOp = builder.create<triton::AddPtrOp>(
+          op.getLoc(), state.source.getType(), state.source, offset);
+      map.map(key, addptrOp.getResult());
+    }
+  }
+
+  for (auto &bodyOp : newOp.getRegion().getOps()) {
+    if (auto forOp = dyn_cast<scf::ForOp>(bodyOp)) {
+      assert(0 && "nested loops currently not supported");
+    }
```
nhat-nguyen, line 839: should we emitRemark and return failure here

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+      auto offset = state.scalar;
+      if (offset.getType().isa<IndexType>()) {
+        offset = builder.create<arith::IndexCastOp>(
+            op.getLoc(), builder.getI32Type(), offset);
+      }
+      auto addptrOp = builder.create<triton::AddPtrOp>(
+          op.getLoc(), state.source.getType(), state.source, offset);
+      map.map(key, addptrOp.getResult());
+    }
```
nhat-nguyen, line 833: similar comment, this logic of handling `state.rank == 0` is new, would be helpful if we have some comments and example snippet here

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+  OpBuilder builder(op);
+
+  // For each of the init arg that we added additional Values in for loop, we
+  // need to add corresponding Values as yield operands. The loop below gathers
+  // PtrSState for those values.
+  SmallVector<PtrSState, 5> initArgState;
+  for (auto [i, v] : llvm::enumerate(op->getOperands())) {
+    // If this operand is not rewritten by forOp, skip
+    auto thisSet = levelToBlockArgIndex.find(level)->second;
```
nhat-nguyen, line 895: nit: `thisSet` doesn't sound too descriptive

include/triton-shared/AnalysisStructured/PtrAnalysis.h
```diff
@@ -0,0 +1,209 @@
...
+
+struct PtrAnalysis {
+  using IndexMapSet = std::map<int, std::set<int>>;
+
+  IndexMapSet levelToBlockArgIndex;
+  int level = 0;
+
+  llvm::SmallDenseMap<Value, PtrSState> knownPtrs;
+
+  IRMapping map;
```
nhat-nguyen, line 74: nit: i think this map should be rename to something a bit more descriptive since it's only used for ptr handling. perhaps something like `ptrMap` would be better

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  }
+
+  if (!ptr.getType().isa<RankedTensorType>()) {
+    op->emitRemark("PtrAnalysis: scalar loadOp will not be rewritten");
+    return failure();
+  }
+
+  ArrayRef<OpFoldResult> dims;
+  MaskSState mstate;
+  Value scalarOther;
```
nhat-nguyen, line 1004: nit: should move this inside the `if (mask)` block

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  }
+
+  if (!ptr.getType().isa<RankedTensorType>()) {
+    op->emitRemark("PtrAnalysis: scalar loadOp will not be rewritten");
+    return failure();
+  }
+
+  ArrayRef<OpFoldResult> dims;
+  MaskSState mstate;
+
```
nhat-nguyen, line 1059: nit: should move inside `if (mask)`

nhat-nguyen:
@haishanzzzz I reviewed the changes and left some comments, overall, I think the approach is good and makes the code much simpler. Love it.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: This scenario is covered by `addptr_scalar_for_2d.mlir`. We still support it just by using slightly different approach. See above.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: Yes exactly. Will add a comment.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+      if (auto makeTPtrOp = mappedV.getDefiningOp<tts::MakeTensorPtrOp>()) {
+        if (visitOperandMakeTPtr(makeTPtrOp, state, op.getLoc(), builder)
+                .failed()) {
+          newInitArgs.push_back(arg);
+          continue;
+        }
+        newInitArgs.push_back(mappedV);
+      } else if (auto addptrOp = mappedV.getDefiningOp<triton::AddPtrOp>()) {
+        assert(!addptrOp.getResult().getType().isa<RankedTensorType>());
+        if (visitOperandAddptr(addptrOp, state, op.getLoc(), builder)
```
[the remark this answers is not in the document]
haishanzzzz: Ah this is again the difference I mentioned above, which is that we don't use `tts.make_tptr` to rewrite `tt.addptr` for scalars. Will add comment.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: We would only hit this statement if it is a pointer that we have successfully analyzed. If it is not pointer related, or if the analyze failed, we would not have pushed into `initArgIndexState`. Does that answer your question?

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: Ah thank you for the suggestion. Will do.

The original code also handles the scenario where we have one argument that is a tensor of index. I don't believe we ever have run into that scenario so I had deleted it.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: Yes, see [this line](https://github.com/llvm/llvm-project/blob/aa23e493f2b363982a472fe38caffc69d907402c/mlir/include/mlir/IR/Builders.h#L227C27-L227C27).

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: I believe this is to account for scenarios where the init arg is a scalar pointer. Will confirm and add comment.

lib/Analysis/OpFoldResultUtils.cpp
```diff
@@ -19,6 +20,32 @@ std::optional<int64_t> getIntAttr(const OpFoldResult ofr) {
   return std::nullopt;
 }
 
+bool hasConstZero(const OpFoldResult ofr) {
+  auto intAttr = getIntAttr(ofr);
+  if (intAttr.has_value()) {
+    if (intAttr.value() == 0) {
```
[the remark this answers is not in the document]
haishanzzzz: That actually doesn't quite do what we want, since we also want to return false right away if intAttr has value but doesn't contain 0.

lib/AnalysisStructured/MaskAnalysis.cpp
haishanzzzz: This is a little tricky on my end. The reason is that I am developing in an internal environment where I only have the local copies of the new files, and if I were to include those memref related functions, it's a lot of changes to the build systems (have to deal with an additional internal one besides the cmake one).

Do you mind if we keep the copies separate for the time being?

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: This is actually no longer needed. Will change the error message.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: This is not obvious but since `dims` is an ArrayRef that can refer to `mstate`'s field, their lifetime need to be the same.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: See above.

lib/AnalysisStructured/PtrAnalysis.cpp
nhat-nguyen: I think multiplying the parent sizes will produce incorrect results though (or rather cases that our formulas won't work with). With the above example again:

```
((2 + tl.arange(0, 3)) % 4) * 2
= [2, 3, 0] * 4
= [8, 12, 0]
```

We finish analyzing `(2 + tl.arange(0, 3)) % 4)` first, at this point we have:
+ offset = 2
+ parent size = 4
+ stride = 1

Then after processing the multiplication, we get:
+ offset = 4
+ parent size = 8
+ stride = 2

So this is when it gets tricky, first the stride is not 1. Our existing offset calculation assumes that the stride is 1 when wrapping around. Also, we will process this as loading from a tensor starting at offset 4 but may wrap around at column 8, when in fact the result `[8, 12, 0]` doesn't represent that; I don't know what it actually means tbh lol. In any case, I think it might be better if we just fail the analysis.

haishanzzzz: Ah I see your concern.

> Then after processing the multiplication, we get parent size = 8
I should not have multiplied the modulos field. That's on me. Thank you for the catch. If it is not multiplied, and if the starting offset is no wrapped around, I think we are still good.

But I agree with you it does gets complicated and . Will remove it for now.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
haishanzzzz: Added explanation on each field on top.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  // AddPtr where both lhs and rhs containing modulo operators not supported
+  if (lhsState.hasModulo() && rhsState.hasModulo()) {
+    op->emitRemark("PtrAnalysis: do not support adding two pointer states "
+                   "that both have modulo");
+    return failure();
+  }
+
+  if (lhsState.hasModulo() || rhsState.hasModulo()) {
+    // visitOperandSplat and visitOperandExpandDims should enforce below
+    assert(lhsState.getRank() <= 2);
```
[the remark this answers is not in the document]
haishanzzzz: I realized that this is to guard the reasoning that `visitOperandSplat and visitOperandExpandDims should enforce` that this rank is true, which they do afaik. Leaving it in since hitting this condition actually indicates that something bad has happened.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+  }
```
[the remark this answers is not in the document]
haishanzzzz: Done.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+  // dealing with modulo:
+  // - If lhs has no modulo, skip
+  // - If rhs has zero offset on dim i, we can just use lhs's modulo
+  // - If i == 0 and rhs is the result of a splat, we will allow the add. This
+  // is because the user may be trying to express adding a constant offset to
+  // increment dim1, but pointer analysis cannot differentiate dim1 vs dim0 in
+  // this case.
+  // - Else, the analysis fail
+
```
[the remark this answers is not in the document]
haishanzzzz: Added example IR for the last case.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: Thanks for the comment. This code is much cleaner now.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: Added comment above this section of code. Your comment actually made me realize that we don't need to create new `tt.addptr` at the beginning of the loop body, only `tts.make_tptr`. I removed dead code and clarified in the comment added.

lib/AnalysisStructured/MaskAnalysis.cpp
manbearian: I'd prefer we reused the old file if possible. Is there a compromise with `#include` or something? Could we annotate OLD/NEW in the file?

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -0,0 +1,58 @@
...
+                scf::SCFDialect, tensor::TensorDialect, triton::TritonDialect,
+                tts::TritonStructuredDialect>();
+  }
+
+  void runOnOperation() override {
+    auto moduleOp = getOperation();
+
+    PtrAnalysis ptrAnalysis;
+    if (ptrAnalysis.rewriteOp(moduleOp).failed()) {
+      moduleOp->emitWarning("PtrAnalysis failed");
```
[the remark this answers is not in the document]
nhat-nguyen: Sounds good. I agree.

lib/AnalysisStructured/PtrAnalysis.cpp
nhat-nguyen: We discussed offline, let's use this approach.

lib/AnalysisStructured/PtrAnalysis.cpp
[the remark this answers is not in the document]
nhat-nguyen: Discussed offline. Current approach is to keep all `tt.addptr` for scalars, these ops will still exist after triton-to-structured. structured-to-memref will need to handle `tt.addptr`.

lib/AnalysisStructured/PtrAnalysis.cpp
nhat-nguyen: This makes sense now. Thank you.

nhat-nguyen:
@haishanzzzz Would you mind updating the description with the decision around keeping `tt.addptr` for scalars after this pass for `structured-to-memref` to process later?

lib/AnalysisStructured/MaskAnalysis.cpp
haishanzzzz: Done. I will just need to be more careful in syncing the files across repos :) Hopefully we can get this rewrite done soon!

haishanzzzz:
> @haishanzzzz Would you mind updating the description with the decision around keeping `tt.addptr` for scalars after this pass for `structured-to-memref` to process later?

Updated the description to include more info on relying on `tt.addptr` for scalar pointers, and that we should expect to see `tt.addptr` at the output of the pass if analysis fails or for scalar pointers.

lib/AnalysisStructured/PtrAnalysis.cpp
haishanzzzz: Closing for now. Please feel free to reopen.

haishanzzzz:
@nhat-nguyen @manbearian Please let me know if there are other things I can do before closing this PR.

nhat-nguyen (approved):
@haishanzzzz   This looks good to me. Thank you for addressing all the comments. You just need to update this with the main branch, once done we will be able to land this.

PR #86, opened 2024-01-18
Add support for tt.extern_elementwise conversion on atan2 and sin operations

aobolensk:
Add extensible converter of tt.extern_elementwise operations invoked on built-in library calls. At the first stage, add conversion to math.atan2 and math.sin MLIR operations


nhat-nguyen (approved):
@aobolensk Looks good to me. Thank you! 😄

PR #87, opened 2024-01-19
Update CPU backend to work with latest triton

nhat-nguyen:
This PR updates the reference CPU backend using the triton-shared middle-layer to use the latest triton 3rd party integration.

## Notable changes
+ `backend/driver.py` and `backend/compiler.py` implement the compiler and driver interfaces required to work with triton.
+ `triton-shared` now includes `triton` as its submodule. The submodule currently points to https://github.com/openai/triton/pull/3007 and will be updated to use the `main` branch once the linked PR is merged.
+ to let triton know we're using the CPU backend, we need to do `triton.runtime.set_active(CPUBackend())` -- see the updated `python/examples` folder for details.

## Other fixes
+ Fixes in `BufferizableOpInterface` due to llvm update in triton main.
+ Lit tests updated due to the above change.
+ previously we could build triton using `python setup.py build`, but that is no longer working after the recent update. We have to do `python3 -m pip install --no-build-isolation -vvv '.[tests]'` directly.


.github/workflows/test-plugin.yml
```diff
@@ -57,26 +44,25 @@ jobs:
...
         python3 -m pip install --upgrade pip
-        python3 -m pip install cmake==3.24
-        python3 -m pip install ninja
-        python3 -m pip uninstall -y triton
-        python3 setup.py build
-        python3 -m pip install --no-build-isolation -vvv '.[tests]'
+        python3 -m pip install cmake==3.24 ninja pytest-xdist
+        sudo apt-get update -y
+        sudo apt-get install -y ccache clang lld
+        export TRITON_PLUGIN_DIRS="${GITHUB_WORKSPACE}/triton_shared"
```
manbearian, line 59: Is there a reason you export this variable but put the others on the python build command?

manbearian (approved):

.github/workflows/test-plugin.yml
nhat-nguyen: I copied the build command from triton main, and since the line is already long, I just put this in another line instead. Also this variable is special so I want others to easily see it.

PR #89, opened 2024-01-20
Populate tt.extern_elementwise conversion with math functions

aobolensk:
Add fabs, sin, cos, tan, log, log10, log1p, exp, exp2, erf, sqrt, rsqrt, ceil, floor, floor, trunc math dialect functions conversion

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -1744,12 +1744,18 @@ public:
     if (op.getLibname().compare("libdevice") || !op.getPure() ||
         op.getArgs().size() != 2)
       return failure();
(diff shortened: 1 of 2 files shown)


nhat-nguyen:
@aobolensk Thank you! We're in the middle of a rewrite to split the pass into several modular passes. This code got moved to `include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp`. Would appreciate if you could update your branch to latest and move the patterns over to that file.

aobolensk:
> @aobolensk Thank you! We're in the middle of a rewrite to split the pass into several modular passes. This code got moved to `include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp`. Would appreciate if you could update your branch to latest and move the patterns over to that file.

Done

nhat-nguyen (approved):

PR #90, opened 2024-01-22
Support Block Pointer in triton-to-structured 

haishanzzzz:
This PR enhances PtrAnalysis logic to support `tt.make_tensor_ptr` and `tt.advance`.

In addition to supporting the ops, this PR makes the following changes:
- Rename `parent_shape` field in `tts.make_tptr` to `shape` to match naming convention of `tt.make_tensor_ptr`
- Add `order` attribute to `tts.make_tptr` to mirror that from `tt.make_tensor_ptr`. If `order` is present, the result of `tts.make_tptr` represents a block pointer and is a pointer to a tensor; otherwise it represents non-block pointer and is a tensor of pointers.

`order` gives us a way to differentiate block pointer type, so later passes can lower it accordingly.  

## Testing
Tested using the existing block_ptr_advance.mlir.


test/Conversion/TritonToStructured/addptr_reshape_broadcast.mlir
```diff
@@ -36,7 +36,7 @@ module {
 // CHECK:         tt.func @kernel([[PARAM_0_:%.+]]: !tt.ptr<bf16, 1>, [[PARAM_1_:%.+]]: !tt.ptr<bf16, 1>) {
 // CHECK-DAG:       [[CST_6144_:%.+]] = arith.constant 6144 : index
 // CHECK-DAG:       [[CST_6_:%.+]] = arith.constant 6 : index
-// CHECK:           [[VAR_0_:%.+]] = tts.make_tptr [[PARAM_1_]] to sizes: [256, 128], strides: [1, [[CST_6_]]{{.}}, offsets: [512, [[CST_6144_]]{{.}}, parent_sizes: [0, 0] : <bf16, 1> to tensor<256x128x!tt.ptr<bf16, 1>>
+// CHECK:           [[VAR_0_:%.+]] = tts.make_tptr [[PARAM_1_]] to sizes: [256, 128], strides: [1, [[CST_6_]]{{.}}, offsets: [512, [[CST_6_]]144], shape: [0, 0], order: [] : <bf16, 1> to tensor<256x128x!tt.ptr<bf16, 1>>
```
nhat-nguyen, line 39: there's a typo here i think, should be `[[CST_6144_]]` instead of `[[CST_6_]]144]`

test/Conversion/TritonToStructured/use_end_chain.mlir
```diff
@@ -47,7 +47,7 @@ module {
 // CHECK:           [[VAR_5_:%.+]] = tt.broadcast [[VAR_4_]] : (tensor<1x128xi32>) -> tensor<256x128xi32>
 // CHECK:           [[VAR_6_:%.+]] = arith.muli [[VAR_5_]], [[VAR_cst_]] : tensor<256x128xi32>
 // CHECK-DAG:       [[VAR_7_:%.+]] = arith.addi [[VAR_2_]], [[VAR_6_]] : tensor<256x128xi32>
-// CHECK-DAG:       [[VAR_8_:%.+]] = tts.make_tptr [[PARAM_1_]] to sizes: [256, 128], strides: [1, [[CST_6_]]{{.}}, offsets: [512, [[CST_6144_]]{{.}}, parent_sizes: [0, 0] : <bf16, 1> to tensor<256x128x!tt.ptr<bf16, 1>>
+// CHECK-DAG:       [[VAR_8_:%.+]] = tts.make_tptr [[PARAM_1_]] to sizes: [256, 128], strides: [1, [[CST_6_]]{{.}}, offsets: [512, [[CST_6_]]144], shape: [0, 0], order: [] : <bf16, 1> to tensor<256x128x!tt.ptr<bf16, 1>>
```
nhat-nguyen, line 50: typo here too

nhat-nguyen:
I have a question regarding the `order` field. I notice that it's only being used to distinguish between a wraparound tensor and a block pointer tensor. Will this field come into play later (e.g.: in `structured-to-memref`)?

nhat-nguyen:
I think this approach makes sense, we reuse `tts::MakeTensorPtrOp` and store additional info to also represent a block pointer created by `triton::MakeTensorPtrOp`.

`triton::MakeTensorPtrOp` offsets are in terms of rows / columns (no strides multiplied). And because our `tts.make_tptr ` offsets already have respective strides multiplied, we have to scale these offsets when handling `tt.advance`.

My only concern is using the `order` filed to distinguish between a wraparound ptr and block ptr can make the logic a little hard to follow; its name doesn't really convey what it really means as I had to refer back to the official triton tutorial. The same can be said for `sizes` too.

I think we don't need to follow the `triton` naming convention here too closely and can probably just keep `parentSizes` for both cases. `order` can probably be renamed to `storageOrder` or something.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -111,9 +111,13 @@ bool PtrState::hasModulo() const {
 }
 
 bool PtrState::dimHasModulo(uint32_t dim) const {
+  assert(
+      order.size() == 0 &&
```
nhat-nguyen, line 115: nit: order.empty()

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -691,6 +737,65 @@ LogicalResult PtrAnalysis::rewriteAddptrOp(triton::AddPtrOp op) {
...
+    auto mulOp = builder.create<arith::MulIOp>(loc, castOp.getResult(),
+                                               stride.get<Value>());
+    auto addOp =
+        builder.create<arith::AddIOp>(loc, mulOp.getResult(), offsetValue);
+    newOffsets.push_back(addOp.getResult());
+  }
+
+  state.offsets.clear();
+
+  for (auto offset : newOffsets) {
```
nhat-nguyen, line 789: i think this code can be simplified to just `state.offsets = newOffsets`

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
```diff
@@ -47,11 +53,20 @@ void LoadOp::build(OpBuilder &b, OperationState &state, Value ptr,
 
   dispatchIndexOpFoldResults(dims, dynamicDims, staticDims);
 
-  auto ptrTensorType = cast<RankedTensorType>(ptr.getType());
-  auto elemType = cast<triton::PointerType>(ptrTensorType.getElementType())
-                      .getPointeeType();
-  auto resType = RankedTensorType::get(ptrTensorType.getShape(), elemType);
+  auto ptrTensorType = dyn_cast<RankedTensorType>(ptr.getType());
+  auto tensorPtrType = dyn_cast<triton::PointerType>(ptr.getType());
```
nhat-nguyen, line 57: nit: this code is really hard to read due to the very similar naming, can we add a comment for each condition indicating whether we're dealing with blockptr or normal ptr? 😄

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
```diff
@@ -21,23 +21,29 @@ void MakeTensorPtrOp::build(OpBuilder &b, OperationState &state, Value base,
...
-                             staticParentSizes);
+  dispatchIndexOpFoldResults(shape, dynamicShape, staticShape);
 
+  Type resType;
   auto basePtr = cast<triton::PointerType>(base.getType());
   auto elemType = basePtr.getPointeeType();
-  auto resType = RankedTensorType::get(sizes, basePtr);
+  if (order.empty()) {
+    resType = RankedTensorType::get(sizes, basePtr);
+  } else {
```
nhat-nguyen, line 38: nit: similar to the other comment in this file, would be good to have a comment indicating which case we're handling. Although we have a comment in the header file saying `order.empty()` means `block ptr`, but it's really not clear on first glance.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -691,6 +737,65 @@ LogicalResult PtrAnalysis::rewriteAddptrOp(triton::AddPtrOp op) {
...
+LogicalResult PtrAnalysis::rewriteAdvanceOp(triton::AdvanceOp op) {
+  OpBuilder builder(op);
+  auto loc = op.getLoc();
+
+  PtrState state;
+  if (visitOperand(op->getOperand(0), state, loc, builder).failed()) {
+    op->emitRemark("PtrAnalysis: Failed to analyze ptr of tt.advance");
+    return failure();
+  }
+  assert(!state.order.empty() && "block pointer should have order field set");
```
nhat-nguyen, line 763: nit: `state.order.empty` can probably be refactored to a helper like `state.isBlockPtr()` or something to make things easier to read

haishanzzzz:
Thanks for the careful review @nhat-nguyen. Will address the nits. 

> We aren't really handling this at all in the old TritonToLinalg pass, so I guess this means we just assume the regular row major storage -- does this match your understanding?

It seems that the prior implementation supports only row major order. See [this](https://github.com/microsoft/triton-shared/blob/703369816542b33e4ba2cea7bbe13a70421d9382/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp#L191C9-L192C74).

> My only concern is using the order filed to distinguish between a wraparound ptr and block ptr can make the logic a little hard to follow; its name doesn't really convey what it really means as I had to refer back to the official triton tutorial. The same can be said for sizes too.
>
> I think we don't need to follow the triton naming convention here too closely and can probably just keep parentSizes for both cases. order can probably be renamed to storageOrder or something.

I am neutral on the naming. The reason I chose the same names is to make it lower the bar to understand these ops/fields, since people who looked at these will definitely have looked at Triton ops. Do you think that makes sense?

nhat-nguyen (approved):

PR #93, opened 2024-02-02
Fix triton-arith-linalg not using the common patterns

nhat-nguyen:
The `triton-arith-to-linalg-pass` wasn't using the common `ConversionPatterns.hpp` file, leading to some inconsistencies in the number of patterns populated.


nhat-nguyen:
@haishanzzzz FYI, I just noticed this because we recently merged a few more patterns in `ConversionPatterns.hpp` but the new pass didn't pick them up.

haishanzzzz:
Thank you for the fix @nhat-nguyen! It turned out that I updated our internal code but didn't do that for triton-shared. This two-copy business is madness :D

manbearian (approved):

PR #94, opened 2024-02-06
Use https for Triton git repo

aaronsm:


nhat-nguyen:
May I ask why we need to change to https? With ssh enabled, I can freely change the branch in the triton submodules and push branches upstream if necessary -- no passwords required after ssh public keys are configured.

aaronsm:
It's a recommendation as this works out of the box without setting up ssh.
This could be addressed in the README I guess by calling out ssh has to be setup.

nhat-nguyen (approved):

PR #98, opened 2024-02-07
Populate tt.extern_elementwise conversion with trigonometric and hyperbolic functions

aobolensk:
Populate conversion pattern implementations with recently added functions to MLIR math dialect
- inverse trigonometric functions (asin, acos, atan)
- hyperbolic functions (sinh, cosh, tanh)
- inverse hyperbolic functions (asinh, acosh, atanh)

PR #100, opened 2024-02-09
Remove library name check for __nv_* intrinsics

aobolensk:
Triton does not set libdevice name for __nv_* intrinsics after https://github.com/openai/triton/pull/2887

- Remove check for libdevice for __nv_* intrinsics
- Remove particular NVIDIA-related libdevice name from the tests

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -1741,8 +1741,7 @@ public:
   matchAndRewrite(triton::ExternElementwiseOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     auto loc = op.getLoc();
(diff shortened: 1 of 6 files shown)


nhat-nguyen:
Thank you! Should we also update the tests to reflect this?

aobolensk:
> Thank you! Should we also update the tests to reflect this?

We have released the limitation here, so library name is not checked, so there is no need to update. But to preserve consistency with what Triton actually emits as of now (could be changed in the future as well), it could be done. WDYT, is it needed?

nhat-nguyen:
Yea I think let's keep the tests consistent with what triton emits, it will be easier to keep track in the future.

aobolensk:
> Yea I think let's keep the tests consistent with what triton emits, it will be easier to keep track in the future.

OK, I'll update after #98 is merged

aobolensk:
Done

nhat-nguyen (approved):

PR #103, opened 2024-02-18
Arm64 workflow runner

danikhan632:


danikhan632:
remade the PR here, should work

nhat-nguyen:
We still have arm-workflow as the target instead of main 😄

danikhan632:
> We still have arm-workflow as the target instead of main 😄

think I misunderstood, so you want me to merge to main?

danikhan632:
> Yeah I think we misunderstood. My question in the other thread was to confirm if you wanted to merge to main or the arm-workflow branch. If we want to merge to the arm-workflow, I can help with updating the arm-workflow branch so your PR's diffs won't be too overwhelming.

I think merging to main might be easier then, I know there are memory issues with 1ES crashing when building triton but we can try main

.github/workflows/test-plugin.yml
```diff
@@ -83,3 +83,76 @@ jobs:
...
+        echo "PATH=${HOME}/.local/bin:${PATH}" >> "${GITHUB_ENV}"
+    - name: Check pre-commit
+      working-directory: triton_shared/triton
+      run: |
+        python3 -m pip install --upgrade pre-commit
+        python3 -m pre_commit run --all-files --verbose
+    - name: Edit setup for arm64
+      working-directory: triton_shared/triton
+      run: |
+        sed -i 's/arch = {"x86_64": "64", "arm64": "aarch64"}/arch = {"x86_64": "64", "arm64": "aarch64", "aarch64":"aarch64"}/' "python/setup.py"
```
aaronsm, line 126: Put up a PR for Triton to change setup.py

.github/workflows/test-plugin.yml
```diff
@@ -83,3 +83,76 @@ jobs:
...
+        sed -i 's|https://tritonlang.blob.core.windows.net/llvm-builds/|https://storage.googleapis.com/compiled-blob/|g' "python/setup.py"
```
aaronsm, line 127: Is this needed after your PR to xcompile the LLVM binaries?

.github/workflows/test-plugin.yml
danikhan632: I believe I already did and it's merged see https://github.com/openai/triton/pull/2999

Not sure if we are still using the custom branch of triton, if we're not then I can delete this line

.github/workflows/test-plugin.yml
danikhan632: I think so until the upstream updates the LLVM version. Although I updated it, don't think it was ram

aaronsm: The arm64 ubuntu build is running and creating the tgz file.

https://github.com/openai/triton/actions/runs/7960036148/job/21728181209

/home/runner/work/triton/triton/llvm-3e77871c-ubuntu-arm64.tar.gz

The job name says Ubuntu 20.04 and could be updated to say Ubuntu 20.04 Arm64.

danikhan632: looks like the upstream cross compiler isnt working need to fix

https://github.com/openai/triton/pull/3134#issuecomment-1957239635
unsure why since

aaronsm: Getting close! I was able to build Triton/Triton-shared with the new tgz on Ubuntu Arm64.

aaronsm: Looks like OAI updated the llvm binaries for arm64 and you can undo the url change

danikhan632: looks like the entire llvm build was cancelled from some odd macos x64 failure. Don't think my PR would have caused that

danikhan632: I believe the bucket has been updated and should be able to be re-ran

danikhan632:
the upstream binaries are updated, can this be re-run?

aaronsm:
> the upstream binaries are updated, can this be re-run?

Sure. The tests will probably fail because the triple is defaulting to X64. We need OAI to update the LLVM binaries one more time with my last change.

aaronsm:
@danikhan632 I think the file needs to be formatted 

https://github.com/microsoft/triton-shared/actions/runs/8086480662/job/22143942797?pr=103

danikhan632:
> @danikhan632 I think the file needs to be formatted 
> 
> 
> 
> https://github.com/microsoft/triton-shared/actions/runs/8086480662/job/22143942797?pr=103

Think it might have been the line spacing

aaronsm (approved):

PR #119, opened 2024-03-25
Update SUPPORT.md

nhat-nguyen:


manbearian:
is this fixing newlines?

nhat-nguyen:
Yeah this is the only file that produced this warning for me:

[LF will be replaced by CRLF in git](https://stackoverflow.com/questions/5834014/lf-will-be-replaced-by-crlf-in-git-what-is-that-and-is-it-important)

manbearian (approved):

PR #120, opened 2024-03-26
Use `add_triton_library` and allow optionally building the CPU backend

nhat-nguyen:
+ Use `add_triton_library` for all triton-shared passes and libs to be consistent with triton. This also helps potential issues when integrating with other backends because `add_mlir_library` exports the targets (and therefore their dependencies) by default, but triton does not export `TritonIR` and other passes.
+ Allow optionally build the cpu backend, this is to account for cases where another backend just wants to use `triton-shared` as a middle-layer and does not need the CPU backend to be installed.
+ Fix lambda capturing the context pointer by reference.
+ Add `get_codegen_implementation` to the CPUBackend, this was added recently in triton upstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,6 @@ add_subdirectory(lib)
 add_subdirectory(test)
 add_subdirectory(tools)
 
-if(TRITON_SHARED_BUILD_CPU_BACKEND)
+if (TRITON_SHARED_BUILD_CPU_BACKEND)
     add_triton_plugin(TritonShared ${CMAKE_CURRENT_SOURCE_DIR}/triton_shared.cc LINK_LIBS TritonSharedAnalysis TritonToLinalg TritonTilingExtIR)
 endif()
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -7,4 +9,7 @@ add_subdirectory(include)
 add_subdirectory(lib)
 add_subdirectory(test)
 add_subdirectory(tools)
-add_triton_plugin(TritonShared ${CMAKE_CURRENT_SOURCE_DIR}/triton_shared.cc LINK_LIBS TritonSharedAnalysis TritonToLinalg TritonTilingExtIR)
+
+if(TRITON_SHARED_BUILD_CPU_BACKEND)
+    add_triton_plugin(TritonShared ${CMAKE_CURRENT_SOURCE_DIR}/triton_shared.cc LINK_LIBS TritonSharedAnalysis TritonToLinalg TritonTilingExtIR)
```
manbearian, line 14: formatting? i don't really care about cmake formatting, but not sure if you meant to do this on purpose.

manbearian (approved):

manbearian (approved):

PR #123, opened 2024-04-01
[StructuredToMemref] Only use `memref.extract_strided_metadata` in loops

nhat-nguyen:
## Intro
We currently lower a sequence of `tt.addptr` for scalars using `memref.extract_strided_metadata` and `memref.reinterpret_cast`. To improve codegen in certain cases, this PR changes the lowering to just using `memref.reinterpret_cast` for `tt.addptr` outside of loops. For `tt.addptr` inside loops, `memref.extract_strided_metadata` is still used to avoid having to carry the pointer offset in the loops' iter args.
 
## Background
In `triton-to-linalg-experimental`, when lowering `%new_ptr = tt.addptr %base_ptr %add_offset` for scalars, `%base_ptr` is lowered to a `memref` using a `memref.reinterpret_cast` op with a certain offset `%curr_offset`. If `%ptr` comes from the function arguments, we assume that the `%curr_offset` is `0`. Then `%new_ptr` is lowered to another `memref` using `memref.reinterpret_cast` with offset `%curr_offset` + `%add_offset`.

With this approach, each `memref.reinterpret_cast` that's lowered from a `tt.addptr` has to know the current offset of `%base_ptr` and perform a pointer addition to get the new offset.

`memref.extract_strided_metadata` can be used to extract the offset from a memref. Then the pointer addition can be done by a simple `arith.addi`.

Doing so is particularly helpful when lowering `tt.addptr` in loops, because we don't have to carry the buffer offsets as induction variables -- the offsets can be retrieved just by calling `memref.extract_strided_metadata`.

## Improvements
With the current approach, the generated code might not be optimal and prevents certain optimizations through canonicalization from kicking in:

```
%0 = memref.reinterpret_cast %arg0 with offset 0

%base, %offset, %size, %stride = memref.extract_strided_metadata %0 : memref<1xf32>, index, index, index

%new_offset = arith.addi %offset %add_offset
%1 = memref.reinterpret_cast %base with offset %new_offset
```

this can actually be simplified to:

```
%1 = memref.reinterpret_cast %base with offset %add_offset
```

When lowering `tt.addptr` not in any loops, to get the current offset the buffer, we can look at the defining op of the buffer which must be a `memref.reinterpret_cast`. So, in the example, we will end up with the following code:

```
%0 = memref.reinterpret_cast %arg0 with offset 0
%new_offset = arith.addi %offset %add_offset
%1 = memref.reinterpret_cast %base with offset %new_offset
```

After canonicalization, we will get:

```
%1 = memref.reinterpret_cast %base with offset %add_offset
```

### DIFF
--- a/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
+++ b/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
@@ -764,10 +764,29 @@ struct ScalarAddptrConverter : public OpConversionPattern<triton::AddPtrOp> {
     auto ptr = adaptor.getPtr();
     auto definingOp = ptr.getDefiningOp();
 
+    // The defining op of `ptr`, if it exists, must always be a reinterpret_cast
+    // because:
+    //  - `ptr` can only come from either a) the result of another tt.addptr,
+    //     b) directly from the function argument, or c) from a block iter arg.
+    //
+    // - for case a), the conversion always replaces tt.addptr with a
+    //   reinterpret_cast. So the defining op of adaptor.getPtr() must be
+    //   a reinterpret_cast.
+    //
+    // - for case b), before applying the conversion, we have already converted
+    //   the function signature to use arguments of memref type using
+    //   unrealized_cast. And these unrealized_cast are converted to
+    //   reinterpret_cast of offset 0 and size 1 in UnrealizedCastConverter.
+    //   So the defining op of adaptor.getPtr() again must be a
+    //   reinterpret_cast.
+    //
+    // - for case c), if the defining op doesn't exist, the tt.addptr op must be
+    //   in a loop.
     if (definingOp) {
       auto reinterpretCast = cast<memref::ReinterpretCastOp>(definingOp);
       auto base = reinterpretCast.getSource();
       auto offsets = reinterpretCast.getMixedOffsets();
+      // Size must always be 1 because we're dealing with scalars.
       assert(offsets.size() == 1);
       Value offset = ofrToIndexValue(offsets[0], loc, rewriter);
 
@@ -779,7 +798,6 @@ struct ScalarAddptrConverter : public OpConversionPattern<triton::AddPtrOp> {
           ArrayRef<OpFoldResult>{rewriter.getIndexAttr(1)} /*strides*/);
 
       rewriter.replaceOp(op, castOp.getResult());
-
     } else {
       // This addptr is being used as an iter-arg in a loop. We use
       // memref.extract_strided_metadata to avoid having to carry on the offset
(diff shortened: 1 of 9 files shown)


lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -762,19 +761,43 @@ struct ScalarAddptrConverter : public OpConversionPattern<triton::AddPtrOp> {
...
-    auto base = extractMetadataOp.getBaseBuffer();
-    auto offset = extractMetadataOp.getOffset();
+    auto ptr = adaptor.getPtr();
+    auto definingOp = ptr.getDefiningOp();
+
+    if (definingOp) {
+      auto reinterpretCast = cast<memref::ReinterpretCastOp>(definingOp);
+      auto base = reinterpretCast.getSource();
+      auto offsets = reinterpretCast.getMixedOffsets();
+      assert(offsets.size() == 1);
```
manbearian, line 771: is this an invariant or should you return failure()  here if offsets != 1

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -762,19 +761,43 @@ struct ScalarAddptrConverter : public OpConversionPattern<triton::AddPtrOp> {
         cast<triton::PointerType>(op.getPtr().getType()).getPointeeType();
     auto memrefType = MemRefType::get({1}, elemType, layout);
 
-    auto extractMetadataOp =
-        rewriter.create<memref::ExtractStridedMetadataOp>(loc, ptr);
-    auto base = extractMetadataOp.getBaseBuffer();
-    auto offset = extractMetadataOp.getOffset();
+    auto ptr = adaptor.getPtr();
+    auto definingOp = ptr.getDefiningOp();
```
manbearian, line 765: The code sequence you have here ends up asserting that the defining op is ReinterpretCastOp. Is that known to be true? is the else block the correct place to go if it is not, rather than an assert?

manbearian (approved):
Looks good, i left a few comments regarding assertions vs. failing the transform.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
nhat-nguyen: It is an invariant because the reinterpret_casts are always created by us. And since we're working with scalars only, the offset should always have size == 1

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
nhat-nguyen: Yeah this is known to be true. I have added a comment explaining why! Thanks for the comment as I realize this isn't clear and might confuse people.

manbearian (approved):

PR #131, opened 2024-04-24
Use structured data class for CPU backend target

### LINKED ISSUE #130: Nightly Build Failure 2024-04-22
cc @microsoft/aifx-compilers
              The nightly build has failed. See: https://github.com/microsoft/triton-shared/actions/runs/8782543761

nhat-nguyen:
Update the sample CPU backend to use structured data class (`GPUTarget`) following https://github.com/openai/triton/commit/5162346487b3e3ebc062d9697429bafad25f22f6.

Fixes #130

### DIFF
--- a/backend/compiler.py
+++ b/backend/compiler.py
@@ -1,4 +1,4 @@
-from triton.backends.compiler import BaseBackend
+from triton.backends.compiler import BaseBackend, GPUTarget
 from triton._C.libtriton import ir, passes
 from dataclasses import dataclass
 from typing import Any, Tuple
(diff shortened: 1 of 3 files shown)


manbearian (approved):

zhaoshiz (commented):
I think this works for now but we'll need to expend Triton's GPUTarget class to create CPU backend variants like x86_64+avx512 or armv8a+sme

nhat-nguyen:
Yeah this is not ideal but is good to start. It's probably best to have a base class `TritonTarget` and let `GPUTarget` and `CPUTarget` inherit from it instead.

PR #135, opened 2024-05-13
Add documentation on how to integrate `triton-shared` with existing triton compilers

nhat-nguyen:
- Add documentation on how to integrate `triton-shared` with existing triton compilers (under `INTEGRATIONS.md`)
- Move the conversion implementation details to `INTERNALS.md`


triton
robbie-vanderzee-microsoft, line 1: Potentially should be separated from this change.

robbie-vanderzee-microsoft (changes requested):
- The instructions are clear and well written. Any adjustments are mostly semantics, and not important.

- I suggest removing the `triton` `submodule` update from this change for regression reasons. Aside from this, I approve this change. 

> [!WARNING]  
> I have not run through them to test integration from scratch; however, I can do so if necessary. As such, my approval is only for the way in which it is written.

manbearian (approved):

PR #148, opened 2024-07-17
User-friendly way to initialize the submodule

parsifal-47:
Original line from the documentation:

```
git clone --recurse-submodules https://github.com/microsoft/triton-shared.git triton_shared
```
uses https to clone triton_shared, while "git@" is used explicitly by the submodule configuration which makes it fail if "git@" is not configured properly, for example, no key added.

Removing the protocol and changing the url to relative resolves this issue.

Thank you!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "triton"]
 	path = triton
-	url = ../../triton-lang/triton.git
+	url = https://github.com/triton-lang/triton.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "triton"]
 	path = triton
-	url = git@github.com:triton-lang/triton.git
+	url = ../../triton-lang/triton.git


nhat-nguyen:
Thanks for taking the time to find issues in the documentation. Would changing the url to use https instead work? I'd prefer that approach over the relative url, mostly because it is not immediately clear how this works without having read the git submodule `man` page:

```
<repository> is the URL of the new submodule's origin repository.
This may be either an absolute URL, or (if it begins with ./ or ../), the location relative to the superproject's origin repository.
```

Also this doesn't work if the project's origin is different from github (say we host the repository on another provider).

parsifal-47:
> Thanks for taking the time to find issues in the documentation. Would changing the url to use https instead work? I'd prefer that approach over the relative url, mostly because it is not immediately clear how this works without having read the git submodule `man` page:
> 
> ```
> <repository> is the URL of the new submodule's origin repository.
> This may be either an absolute URL, or (if it begins with ./ or ../), the location relative to the superproject's origin repository.
> ```
> 
> Also this doesn't work if the project's origin is different from github (say we host the repository on another provider).

1. yes, if we change to "https" that should work better and will be more consistent with current documentation.
2. I think relative references are very common, here is the quote from git documentation:
"Since the URL in the .gitmodules file is what other people will first try to clone/fetch from, make sure to use a URL that they can access if possible. For example, .... When applicable, a relative URL can be helpful."
https://git-scm.com/book/en/v2/Git-Tools-Submodules

Anyway, I am ok with changing it to "https".
Thank you!

nhat-nguyen:
I think let's use the absolute https url since there might be cases where the users have their own internal mirror of triton-shared that isn't hosted on github. The absolute url will still work without requiring the triton repo to also be mirrored.

parsifal-47:
updated, please take a look, thank you!

nhat-nguyen (approved):
Thank you!

PR #149, opened 2024-07-17
Adding a way to run tests

parsifal-47:
It is useful to have a way to run tests documented,
please feel free to correct me or suggest a different solution to do that.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,11 @@ sudo apt-get install -y ccache clang lld
 TRITON_BUILD_WITH_CLANG_LLD=true TRITON_BUILD_WITH_CCACHE=true python3 -m pip install --no-build-isolation -vvv '.[tests]'
 ```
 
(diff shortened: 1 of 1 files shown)


parsifal-47:
I have found that there is a way to run python/examples as tests, but tests that are in "test/" are probably runnable with "lit test"

nhat-nguyen:
Hi @parsifal-47, there are two kind of tests that we have:

- runtime execution tests that run on CPU (`python/examples`)
- lit tests (`test`)

Both run on each PR and nightly (every night with the latest triton commit from upstream), please refer to this file for more information: https://github.com/microsoft/triton-shared/blob/main/.github/workflows/test-plugin.yml

If the nightly pipeline fails, an issue is automatically created.

parsifal-47:
Thank you for the explanation!

PR #151, opened 2024-07-19
Add nested loops support to PtrAnalysis

nhat-nguyen:
## Introduction

This PR adds nested loops support to `PtrAnalysis`. The implementation does not have any limits to the number of nesting levels in any given loops. There should be no functional differences except slight changes in the passing order of pointer offsets and strides into `scf.for`'s init-args in certain cases.
## Motivation

The current implementation of `PtrAnalysis` does not support multiple levels of loops. There is support for perfectly nested loop internally at Meta (i.e.: loops that only have body in their inner-most level), but it has not been upstreamed yet. As we aim to support more triton programs, it would be beneficial to have a more generalized approach to lowering triton programs that can contain even non-perfectly nested loops.
## Background

The current implementation follows these (simplified) steps when rewriting loops:

1. create a copy of the existing `scf.for` op; for every pointer that appears in the `scf.for` op's iter-args, create additional arguments in the loop that describe the pointer's offsets and strides.
2. the `PtrState` of the pointers carried in the `scf.for` are then:
    + used to initialize the loop's init-args (the offsets and strides),
    + used to prepopulate the `PtrState` for the pointers in the loop's iter-args
        + the `PtrState` for a pointer in a loop's iter-arg is the same as its init-arg counterpart, except that its offsets and strides point to the corresponding iter-args created in step 1.
        + this state is used in step 3 during visits of the new loop's inner ops; in essence, before we visit a loop level, we have to prepare the `PtrState` for all of its iter-arg.
3. visit the new op's body and rewrite the inner ops.
4. update the new `scf.for`'s yield op to return the additionally created offsets and strides for each pointer.

Unfortunately, scaling to multiple levels of loop nest using the existing approach is particularly difficult for a number of reasons. There are a lot of the complexity in having to *manually* add the pointers' offsets and strides to the loop arguments. Below are some notable limitations that were considered.

#### 1. Many data structures to maintain

There are several data structures that keep track of which `PtrState` corresponds to which loop argument: some are local to a level of nesting, some are global (`levelToBlockArgIndex`). Running the same algorithm recursively will mean more complexity and make the whole process harder to follow.

#### 2. Rewriting yield ops is complex

Rewriting the yield op for one level of `scf.for` requires the `PtrStates` for the `yield` operands to have already been populated; this requirement means that we need to have already processed the current loop body before rewriting the `yield`.

So for nested loops, we need to have already visited all the inner loops before being able to process the outer loops' `yield` ops. But remember that for each level of loop nest, we first need to have already processed the outer's loop body in order to have the loop's iter-args' `PtrStates` ready in step 2. described above. As a result, our rewrite process will require rewriting yield in a post-order fashion, while the setup of the `PtrStates` will be in a pre-order fashion. While all of this is totally possible,  the implementation will be hard to follow and understand.

#### 3. Deep cloning `scf.for` for every loop level is inefficient

For each loop level, we have to clone all of its children in order to accommodate the new init-args. With the introduction of multiple loop levels, we will end up having multiple copies of the loops, each with the immediate outer level removed which is quite inefficient. Real world scenarios might not have deeply nested loops, but loop bodies with large number of ops inside will still incur some cost.

## New solution

Because a lot of the complexity in support multiple levels of loop nests comes from the manual rewrites of the `scf.for` and `scf.yield` to add the pointer offsets and strides from `PtrState`, I have decoupled this step from the rest of the main `PtrAnalysis` pass in hope of making everything more understandable and maintainable. Here's a high level overview of the new flow:

#### 1. `PtrAnalysis` pre-processing pass

Rewrites of `scf.for` to add pointer offsets and strides from `PtrState` is done in a preprocessing pass. This preprocessing pass maps all uses of every pointer that is used in a `scf.for` loop (type `tt.ptr` or `tensor<tt.ptr>`) to a sequence of `tt.ptr, offset_0, offset_1,... ,stride_0, stride_1,...` by using a new op `tts.get_structured_state`. Because the `PtrStates` do not exist yet, these ops serve as placeholders for us to eventually insert the correct information once we have computed the `PtrState`. In summary, for each pointer value, we create its corresponding strides and offsets that will be used later on in the rewrite process.

This conversion is done using the `1-N converter` provided by `mlir`. This pass can be run individually by using `triton-shared-opt --triton-to-structured="run-prepass-only=true"`. Notice that the type conversion we're doing is recursive without a base case (pointer type converts to a sequence of pointer type followed by index types for offsets and strides), we first have to convert the pointer types to a tuple of pointer type and index first before finally "decomposing" the tuple back to the original types. Comments in the pass will explain the process in more details.

Running the preprocessing step with the input:

```mlir
%11 = tt.addptr %10, %9 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%20 = tt.addptr %18, %19 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%23:2 = scf.for %arg8 = %c0_i32 to %c2_i32 step %c1_i32 iter_args(%arg9 = %11, %arg10 = %20) -> (tensor<2x2x!tt.ptr<f32>>, tensor<2x2x!tt.ptr<f32>>)  : i32 {
  %24 = tt.load %arg9 : tensor<2x2x!tt.ptr<f32>>
  tt.store %arg10, %24 : tensor<2x2x!tt.ptr<f32>>
  %25 = tt.addptr %arg9, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
  %26 = tt.addptr %arg10, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
  scf.yield %25, %26 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2x!tt.ptr<f32>>
}
```

We get:

```mlir
%11 = tt.addptr %10, %9 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%20 = tt.addptr %18, %19 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%23:5 = "tts.get_structured_state"(%11) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
%24:5 = "tts.get_structured_state"(%20) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
%25:10 = scf.for %arg8 = %c0_i32 to %c2_i32 step %c1_i32 iter_args(%arg9 = %23#0, %arg10 = %23#1, %arg11 = %23#2, %arg12 = %23#3, %arg13 = %23#4, %arg14 = %24#0, %arg15 = %24#1, %arg16 = %24#2, %arg17 = %24#3, %arg18 = %24#4) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index, tensor<2x2x!tt.ptr<f32>>, index, index, index, index)  : i32 {
    %26 = tt.load %arg9 : tensor<2x2x!tt.ptr<f32>>
    tt.store %arg14, %26 : tensor<2x2x!tt.ptr<f32>>
    %27 = tt.addptr %arg9, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
    %28 = tt.addptr %arg14, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
    %29:5 = "tts.get_structured_state"(%27) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
    %30:5 = "tts.get_structured_state"(%28) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
    scf.yield %29#0, %29#1, %29#2, %29#3, %29#4, %30#0, %30#1, %30#2, %30#3, %30#4 : tensor<2x2x!tt.ptr<f32>>, index, index, index, index, tensor<2x2x!tt.ptr<f32>>, index, index, index, index
}
```

#### 2. PtrAnalysis with recursive `rewriteForOp`

The PtrAnalysis pass is similar to before, with the following notable changes:
+ `rewriteForOp` now recursively visits all inner ops and is more simplified.
+ `rewriteYieldOp` is removed because the `yields` are already rewritten to the correct form in the pre-processing pass.
+ because multiple levels of loops are now supported, `visitOperand` now has to support values returned by scf.for (e.g.: pointers returned by an inner loop are used by outer loops)

#### 3. Process `tts.get_structured_state`

Now that all of the PtrStates have been populated by the PtrAnalysis pass, we can revisit all the `tts.get_structured_state` placeholders to insert the correct initialization ops for the offsets and strides. If the analysis is successful, all of the `tts.get_structured_state` ops are expected to be removed at the end. The implementation for this step is basically the same as the initial step where we set up the init-arg for the scf.for in the existing implementation.

## Testing

Multiple lit tests are added with various scenarios:
- tests for the preprocessing steps
- tests for the triton-to-structured-pass
- tests for the end-to-end triton-to-linalg pass
- impractical tests with many levels of nesting and multiple loops in the same level

End-to-end tests running on the CPU backend:
+ 4 tests under `python/examples/test_nested_loops.py` running on the CPU backend
    + the results are compared against triton running with CUDA


parsifal-47:
I know this is work in progress, but please consider adding new tests to the integration

nhat-nguyen:
@parsifal-47 I'm still in the process of ironing things out, but there are a few tests in the draft PR already -- both lit and integration tests.

PR #152, opened 2024-07-23
Implement support for multiple levels of loop nest

nhat-nguyen:
## Introduction

This PR adds nested loops support to `PtrAnalysis`. The implementation does not have any limits to the number of nesting levels in any given loops. There should be no functional differences except slight changes in the passing order of pointer offsets and strides into `scf.for`'s init-args in certain cases.

## Motivation

The current implementation of `PtrAnalysis` does not support multiple levels of loops. There is support for perfectly nested loop internally at Meta (i.e.: loops that only have body in their inner-most level), but it has not been upstreamed yet. As we aim to support more triton programs, it would be beneficial to have a more generalized approach to lowering triton programs that can contain even non-perfectly nested loops.

## Background

The current implementation follows these (simplified) steps when rewriting loops:

1. create a copy of the existing `scf.for` op; for every pointer that appears in the `scf.for` op's iter-args, create additional arguments in the loop that describe the pointer's offsets and strides.
2. the `PtrState` of the pointers carried in the `scf.for` are then:
    + used to initialize the loop's init-args (the offsets and strides),
    + used to prepopulate the `PtrState` for the pointers in the loop's iter-args
        + the `PtrState` for a pointer in a loop's iter-arg is the same as its init-arg counterpart, except that its offsets and strides point to the corresponding iter-args created in step 1.
        + this state is used in step 3 during visits of the new loop's inner ops; in essence, before we visit a loop level, we have to prepare the `PtrState` for all of its iter-arg.
3. visit the new op's body and rewrite the inner ops.
4. update the new `scf.for`'s yield op to return the additionally created offsets and strides for each pointer.

Unfortunately, scaling to multiple levels of loop nest using the existing approach is particularly difficult for a number of reasons. There are a lot of the complexity in having to *manually* add the pointers' offsets and strides to the loop arguments. Below are some notable limitations that were considered.

#### 1. Many data structures to maintain

There are several data structures that keep track of which `PtrState` corresponds to which loop argument: some are local to a level of nesting, some are global (`levelToBlockArgIndex`). Running the same algorithm recursively will mean more complexity and make the whole process harder to follow.

#### 2. Rewriting yield ops is complex

Rewriting the yield op for one level of `scf.for` requires the `PtrStates` for the `yield` operands to have already been populated; this requirement means that we need to have already processed the current loop body before rewriting the `yield`.

So for nested loops, we need to have already visited all the inner loops before being able to process the outer loops' `yield` ops. But remember that for each level of loop nest, we first need to have already processed the outer's loop body in order to have the loop's iter-args' `PtrStates` ready in step 2. described above. As a result, our rewrite process will require rewriting yield in a post-order fashion, while the setup of the `PtrStates` will be in a pre-order fashion. While all of this is totally possible,  the implementation will be hard to follow and understand.

#### 3. Deep cloning `scf.for` for every loop level is inefficient

For each loop level, we have to clone all of its children in order to accommodate the new init-args. With the introduction of multiple loop levels, we will end up having multiple copies of the loops, each with the immediate outer level removed which is quite inefficient. Real world scenarios might not have deeply nested loops, but loop bodies with large number of ops inside will still incur some cost.

## New solution

Because a lot of the complexity in support multiple levels of loop nests comes from the manual rewrites of the `scf.for` and `scf.yield` to add the pointer offsets and strides from `PtrState`, I have decoupled this step from the rest of the main `PtrAnalysis` pass in hope of making everything more understandable and maintainable. Here's a high level overview of the new flow:

#### 1. `PtrAnalysis` pre-processing pass

Rewrites of `scf.for` to add pointer offsets and strides from `PtrState` is done in a preprocessing pass. This preprocessing pass maps all uses of every pointer that is used in a `scf.for` loop (type `tt.ptr` or `tensor<tt.ptr>`) to a sequence of `tt.ptr, offset_0, offset_1,... ,stride_0, stride_1,...` by using a new op `tts.get_structured_state`. Because the `PtrStates` do not exist yet, these ops serve as placeholders for us to eventually insert the correct information once we have computed the `PtrState`. In summary, for each pointer value, we create its corresponding strides and offsets that will be used later on in the rewrite process.

This conversion is done using the `1-N converter` provided by `mlir`. This pass can be run individually by using `triton-shared-opt --triton-to-structured="run-prepass-only=true"`. Notice that the type conversion we're doing is recursive without a base case (pointer type converts to a sequence of pointer type followed by index types for offsets and strides); so we first have to convert the pointer types to a tuple of pointer type and index first before finally "decomposing" the tuple back to the original types. Comments in the pass will explain the process in more details.

Running the preprocessing step with the input:

```mlir
%11 = tt.addptr %10, %9 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%20 = tt.addptr %18, %19 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%23:2 = scf.for %arg8 = %c0_i32 to %c2_i32 step %c1_i32 iter_args(%arg9 = %11, %arg10 = %20) -> (tensor<2x2x!tt.ptr<f32>>, tensor<2x2x!tt.ptr<f32>>)  : i32 {
  %24 = tt.load %arg9 : tensor<2x2x!tt.ptr<f32>>
  tt.store %arg10, %24 : tensor<2x2x!tt.ptr<f32>>
  %25 = tt.addptr %arg9, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
  %26 = tt.addptr %arg10, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
  scf.yield %25, %26 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2x!tt.ptr<f32>>
}
```

We get:

```mlir
%11 = tt.addptr %10, %9 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%20 = tt.addptr %18, %19 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
%23:5 = "tts.get_structured_state"(%11) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
%24:5 = "tts.get_structured_state"(%20) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
%25:10 = scf.for %arg8 = %c0_i32 to %c2_i32 step %c1_i32 iter_args(%arg9 = %23#0, %arg10 = %23#1, %arg11 = %23#2, %arg12 = %23#3, %arg13 = %23#4, %arg14 = %24#0, %arg15 = %24#1, %arg16 = %24#2, %arg17 = %24#3, %arg18 = %24#4) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index, tensor<2x2x!tt.ptr<f32>>, index, index, index, index)  : i32 {
    %26 = tt.load %arg9 : tensor<2x2x!tt.ptr<f32>>
    tt.store %arg14, %26 : tensor<2x2x!tt.ptr<f32>>
    %27 = tt.addptr %arg9, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
    %28 = tt.addptr %arg14, %22 : tensor<2x2x!tt.ptr<f32>>, tensor<2x2xi32>
    %29:5 = "tts.get_structured_state"(%27) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
    %30:5 = "tts.get_structured_state"(%28) : (tensor<2x2x!tt.ptr<f32>>) -> (tensor<2x2x!tt.ptr<f32>>, index, index, index, index)
    scf.yield %29#0, %29#1, %29#2, %29#3, %29#4, %30#0, %30#1, %30#2, %30#3, %30#4 : tensor<2x2x!tt.ptr<f32>>, index, index, index, index, tensor<2x2x!tt.ptr<f32>>, index, index, index, index
}
```

#### 2. PtrAnalysis with recursive `rewriteForOp`

The `PtrAnalysis` pass is similar to before, with the following notable changes:
+ `rewriteForOp` now recursively visits all inner ops and is more simplified.
+ `rewriteYieldOp` is removed because the `yields` are already rewritten to the correct form in the pre-processing pass.
+ because multiple levels of loops are now supported, `visitOperand` now has to support values returned by scf.for (e.g.: pointers returned by an inner loop are used by outer loops)

#### 3. Process `tts.get_structured_state`

Now that all of the `PtrStates` have been populated by the `PtrAnalysis` step, we can revisit all the `tts.get_structured_state` placeholders to insert the correct initialization ops for the offsets and strides. If the analysis is successful, all of the `tts.get_structured_state` ops are expected to be removed at the end. The implementation for this step is basically the same as the initial step where we set up the init-arg for the `scf.for` in the existing implementation.

### Potential differences in ordering of `scf.for` arguments
Because the new approach inserts the pointer offsets and strides next to their corresponding pointers, the ordering of `scf.for` arguments coming out of the `triton-to-structured` pass might differ slightly from the previous implementation. However, this should be fairly rare as the ordering difference only reflects in a few lit tests.

## Testing

Multiple lit tests are added with various scenarios:
- tests for the preprocessing steps
- tests for the triton-to-structured-pass
- tests for the end-to-end triton-to-linalg pass
- impractical tests with many levels of nesting and multiple loops in the same level

End-to-end tests running on the CPU backend:
+ 4 tests under `python/examples/test_nested_loops.py` running on the CPU backend
    + the results are compared against triton running with CUDA


nhat-nguyen:
@haishanzzzz This is the PR for the nested loop support that I mentioned last week. It's ready to be reviewed now. 😄

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -120,6 +120,16 @@ def TTS_MakeTensorPtrOp
...
 }
 
+def TTS_GetStructuredStateOp : TTS_Op<"get_structured_state", [Pure]> {
+  let summary = "Placeholder for the structured pointer states computed during PtrAnalysis.";
+  let description = "Used to pass the offsets and strides to scf.for op to simplify IR rewrites.";
+
+  let arguments = (ins TT_PtrLike:$ptr);
+  let results = (outs Variadic<AnyType>:$outputs);
+  let hasFolder = 0;
+  let hasVerifier = 0;
```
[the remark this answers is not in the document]
red1bluelost: Or if its scalar, just one index type.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -694,6 +726,9 @@ LogicalResult PtrAnalysis::visitOperand(Value operand, PtrState &state,
     }
   }
 
+  if (!operand.getDefiningOp()) {
+  }
```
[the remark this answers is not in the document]
nhat-nguyen: right! thanks for noticing this

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -797,145 +841,141 @@ LogicalResult PtrAnalysis::rewriteAdvanceOp(triton::AdvanceOp op) {
...
-        auto constOp = builder.create<arith::ConstantOp>(
-            op.getLoc(), builder.getIndexAttr(sIntAttr.value()));
-        newInitArgs.push_back(constOp.getResult());
-        state.offsets[j] = constOp.getResult();
-      } else {
-        newInitArgs.push_back(s.get<Value>());
-      }
+PtrState PtrAnalysis::reconcileLoopPtrState(
+    scf::ForOp forOp, size_t iterArgIndex, const PtrState &state,
+    std::function<Value(scf::ForOp op, size_t)> getReplacementVal) {
```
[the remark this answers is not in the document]
nhat-nguyen: ah good to know, i will change to llvm::function_ref

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -953,167 +993,81 @@ LogicalResult PtrAnalysis::rewriteForOp(scf::ForOp op) {
...
-      return failure();
-    }
-  }
-
-  // Update the loop body.
-  if (rewriteOp(newOp).failed()) {
-    newOp->erase();
+  // Recursively rewrite the inner ops
+  if (rewriteOp(op).failed()) {
+    op->erase();
```
haishanzzzz, line 1012: iiuc, this deletes the original for op? Why do we want to do that?

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -41,13 +57,244 @@ class TritonToStructuredPass
...
+  LogicalResult convertToPointerTupleWithOffsetsAndStrides() {
     auto moduleOp = getOperation();
 
+    RewritePatternSet patterns(&getContext());
+
+    auto context = &getContext();
+    OneToNTypeConverter converter;
+    converter.addConversion([](Type type) { return type; });
+
+    // We are doing a 1->1 type conversion here, where a triton pointer type
```
haishanzzzz, line 69: nit, do you mean "1->N"? :)

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -41,13 +57,244 @@ class TritonToStructuredPass
...
+      if (auto ptrType =
+              dyn_cast<triton::PointerType>(tensorType.getElementType())) {
+        SmallVector<Type> tupleTypes{tensorType};
+        // Each tensor of rank k gets k values for its offsets and k values for
+        // its strides, all of which has Index type.
+        tupleTypes.append(tensorType.getRank() * 2, IndexType::get(context));
+        types = SmallVector<Type>{TupleType::get(context, tupleTypes)};
+        return success();
+      }
+      return std::nullopt;
```
haishanzzzz, line 86: nit, `return failure();`

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -41,13 +57,244 @@ class TritonToStructuredPass
...
+    // convert that value to a tuple type.
+    converter.addTargetMaterialization(
+        [](OpBuilder &builder, TypeRange resultTypes, Value input,
+           Location loc) -> std::optional<SmallVector<Value>> {
+          return builder
+              .create<UnrealizedConversionCastOp>(loc, resultTypes, input)
+              ->getResults();
+        });
+
+    scf::populateSCFStructuralOneToNTypeConversions(converter, patterns);
```
haishanzzzz, line 137: This definitely allows us to simplify our code by quite a bit, but I am slightly worried about one specific scenario.

Assume there is an iter arg of a scf.for is a tensor of integer; the loop body uses this iter arg to create a non-block pointer. To support scenarios like this, we may want to visit the use of the iter arg before deciding that we'd want to convert this iter arg to a structured format too. 

I believe we cannot rely on TypeConverter to support such scenarios, since type rewrite cannot be conditional. 

I understand this is non-trivial, but would it be possible for us to keep the manual rewriting framework (or if you know better ways to support such cases)?

manbearian: >would it be possible for us to keep the manual rewriting framework
 i really hope there's a better way to handle this...

include/triton-shared/AnalysisStructured/PtrAnalysis.h
```diff
@@ -189,6 +198,43 @@ struct PtrAnalysis {
...
+
+  // Get the computed PtrState for the forOp's iter-arg at the provided index.
+  LogicalResult getLoopIterArgPtrState(scf::ForOp forOp, size_t index,
+                                       PtrState &state);
+
+  // Get the computed PtrState for the forOp's result at the provided index.
+  LogicalResult getLoopResultPtrState(scf::ForOp forOp, size_t index,
+                                      PtrState &state);
+
+  // This function is internally used by getLoopIterArgPtrState and
```
manbearian, line 213: should this be in the header file if it is "internally used"?

include/triton-shared/AnalysisStructured/PtrAnalysis.h
```diff
@@ -189,6 +198,43 @@ struct PtrAnalysis {
                                           PtrState &state, const Location loc,
                                           OpBuilder &builder);
 
+  // Get the computed PtrState for the forOp's init-arg at the provided index.
+  LogicalResult getLoopInitArgPtrState(scf::ForOp forOp, size_t index,
```
manbearian, line 202: These functions might be rewritten to return `FailureOr<PrtrState>` ?

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
nhat-nguyen: @haishanzzzz @manbearian I believe we can still support the case that you mentioned using this new type converter approach.

The changes applied to the IR in the prepass are reversible by running canonicalization again. What this means is, we can always safely apply the same transformation to *all* tensors of integers to also include the additional offsets and strides *just in case* the tensors are used as part of a pointer arithmetic sequence.

The PtrAnalysis pass can then do some analysis to determine whether a tensor of integer needs to be converted to the structured format. If a structured format is necessary for a given tensor of integer, it means that the offsets and strides included in the loop are used. Otherwise, the offsets and strides are "dead" and can be safely removed by canonicalization after PtrAnalysis finishes. This approach is similar to cases where PtrAnalysis cannot analyze pointers and have to remap back to the "original pointer" -- in such cases, the offsets and strides inserted by the prepass are also removed.

include/triton-shared/AnalysisStructured/PtrAnalysis.h
nhat-nguyen: yeah that is perfect, i will update to use `FailureOr`

lib/AnalysisStructured/PtrAnalysis.cpp
nhat-nguyen: yeah this is my mistake

include/triton-shared/AnalysisStructured/PtrAnalysis.h
nhat-nguyen: I think it's worth putting it together in the header file rather than a static function in the cpp file because its implementation is closely related to the PtrAnalysis pass. I have marked this function as private instead.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -120,6 +120,16 @@ def TTS_MakeTensorPtrOp
   //let hasCanonicalizer = 1;
 }
 
+def TTS_GetStructuredStateOp : TTS_Op<"get_structured_state", [Pure]> {
+  let summary = "Placeholder for the structured pointer states computed during PtrAnalysis.";
+  let description = "Used to pass the offsets and strides to scf.for op to simplify IR rewrites.";
+
+  let arguments = (ins TT_PtrLike:$ptr);
+  let results = (outs Variadic<AnyType>:$outputs);
```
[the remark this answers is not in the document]
nhat-nguyen: Thanks for this. The IR actually looks easier to read with this change too.

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
nhat-nguyen: This is actually intended and caught me off guard too. There's a subtle difference between returning `failure()` and `std::nullopt`. From the [documentation](https://github.com/llvm/llvm-project/blob/3210bce90f8678d52ea67f812bfe81d1c9de7597/mlir/include/mlir/Transforms/DialectConversion.h#L152):

> If `std::nullopt` is returned, the converter is allowed to try another conversion function to perform the conversion.

Say we have type `tensor<4x256xbf16>` which is a `RankedTensorType`. Even though this `RankedTensorType` matches the converter that handles the tuple conversion, we want to keep this type as is because the inner type isn't a pointer. So by returning `failure()`, the TypeConverters will stop trying the remaining converters. In our case, the last type converter which simply returns the same type is skipped. And because the conversion for this type has failed, the whole conversion process is also skipped.

Relevant links to the implementation:
+ https://github.com/llvm/llvm-project/blob/cb5dc1faa8b3702e0d03426ee5dfc5e1b903ec47/mlir/lib/Transforms/Utils/DialectConversion.cpp#L2958
+ https://github.com/llvm/llvm-project/blob/cb5dc1faa8b3702e0d03426ee5dfc5e1b903ec47/mlir/lib/Transforms/Utils/DialectConversion.cpp#L3033

manbearian: worth a comment in the code!

nhat-nguyen:
@haishanzzzz I've cleaned up the code based on yours and others' feedback, would be great if you can take another look at this. Thank you!

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
haishanzzzz: That makes sense. Thank you for the explanation. Would you want to do that within this PR, or should we leave it for later?

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -41,13 +57,244 @@ class TritonToStructuredPass
...
+  // then take {triton_ptr, offset_0, offset_1, ..., stride_0, stride_1,...}.
+  //
+  // The 1->N type conversion is a perfect fit for this transformation.
+  // Unfortunately, we cannot do this is one pass, because the current 1->N
+  // type conversion implementation for scf.for ops doesn't provide us with a
+  // way to detect that a type conversion is recursive. So a triton_ptr type
+  // that gets converted to a {triton_ptr, offset_0, offset_1, ..., stride_0,
+  // stride_1,...} will recursively trigger other conversions.
+  //
+  // To fix this issue, we have to first convert triton_ptr to
```
haishanzzzz, line 277: Thank you for the detailed comment here!

haishanzzzz (approved):
Thank you for the explanation. Sounds good. LGTM.

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
nhat-nguyen: I can do another follow up PR to implement this. Do you happen to have a triton test case handy? Otherwise it's fine too, I can try generating one. 😄

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
```diff
@@ -86,5 +93,81 @@ void StoreOp::build(OpBuilder &b, OperationState &state, Value ptr, Value value,
...
+      getOffsetAndStrideTypes(getContext(), getStructuredPtr().getType());
+
+  if (!expectedOffsetAndStrideTypes.has_value()) {
+    return failure();
+  }
+
+  auto [expectedOffsetTypes, expectedStrideTypes] =
+      *expectedOffsetAndStrideTypes;
+
+  return success(expectedOffsetTypes.size() == getOffsets().size() &&
```
robbie-vanderzee-microsoft, line 108: Does `llvm::equal` not override for the `.size()` type?

robbie-vanderzee-microsoft (approved):
This looks good. Highly appreciate the thorough walkthroughs in the comments. No appreciable adjustment suggestions to provide. 

Testing may benefit from a simpler example, but is sufficient.

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
nhat-nguyen: yeah it is a wrapper for std::equal which doesn't do bound check

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -41,13 +67,247 @@ class TritonToStructuredPass
...
+    }
+
+    if (runPrepassOnly) {
+      return;
+    }
+
+    auto moduleOp = getOperation();
     mlir::tts::PtrAnalysis ptrAnalysis;
-    if (ptrAnalysis.rewriteOp(moduleOp).failed()) {
+    if (failed(ptrAnalysis.rewriteOp(moduleOp))) {
```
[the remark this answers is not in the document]
nhat-nguyen: there's no difference in functionality, just convention 😄

PR #158, opened 2024-08-02
Adding kernel code to the hash

parsifal-47:
Only the wrapper was used to evaluate the key for cache_manager, as a result, all kernels with same name and arguments did not recompile which caused a lot of simple tests to fail. For example, binary operations, swapping + for - inside the kernel resulted in the first compiled to be used.

With the change all binary tests are passing, before most of them failed because of improper caching.


parsifal-47:
@nhat-nguyen please take a look when you have a chance, thank you!

nhat-nguyen (approved):
Sweet. Thank you!

nhat-nguyen:
Just some thoughts on our testing strategy (we don't need to address it here). It would be great if we can somehow reuse the tests from the triton repository without having to duplicate them. Not a huge deal right now given we're only beginning to add the tests. But at one point it may become a bit hard to manage and sync with the upstream ones. Anyway, thank you again for this patch!

parsifal-47:
> Just some thoughts on our testing strategy (we don't need to address it here). It would be great if we can somehow reuse the tests from the triton repository without having to duplicate them. Not a huge deal right now given we're only beginning to add the tests. But at one point it may become a bit hard to manage and sync with the upstream ones. Anyway, thank you again for this patch!

I agree, I was also thinking about that, I think I can write conftest.py for pytest in a way that we can reuse embed CPUBackend everywhere, the only thing is `@pytest.mark.interpreter` fixture, not sure yet if it would be possible to remove it. Anyway, if most of the core tests are passing now, I will look how to remove these copy-pasted files and use from triton.

parsifal-47:
I briefly checked, this decorator is relatively easy to neutralize, there are some other minor modifications in the code, but overall looks doable. In the worst case, I would suggest modifications in the triton repo to make these tests reusable. Anyway, will address that in subsequent PRs, thank you!

parsifal-47:
@nhat-nguyen since this PR wasn't merged yet, I decided to pile deduplication effort on top of it, let me know if I need to separate it. Here is the summary of the recent changes:

1. I put all CPUBackend custom logic into pytest configuration file, removed it from all the tests
2. Made previously copy-pasted files to be symlinks to triton repository
3. Added exclude list for coretests, it turned out that roughly half (by name) of them are still not passing, the plan is to expand the support and remove them from the list once the fixes are available.
4. Since device="cpu" is a fixture now, removed hardcoded value from the sources
5. Slightly refactored test sources since I touched them, removed debug prints

Current stats: 2774 passed, 9733 skipped. A lot of these skipped tests are variants.

nhat-nguyen:
Looks like these new tests exposed some missing symbols in the pipeline:

```
E   ImportError: /home/runner/.triton/cache/bd3cd9ef8032e186d41db6c8267a59d0/__triton_shared_ref_cpu_kernel_launcher.so: undefined symbol: __truncsfbf2
```

parsifal-47:
> Looks like these new tests exposed some missing symbols in the pipeline:
> 
> ```
> E   ImportError: /home/runner/.triton/cache/bd3cd9ef8032e186d41db6c8267a59d0/__triton_shared_ref_cpu_kernel_launcher.so: undefined symbol: __truncsfbf2
> ```

I see, bfloat16 doesn't work for some reason, but it works for me, interesting. Let me see if I can mask these specific testcases that operate on bfloat16. Thank you!

parsifal-47:
I just pushed the fix, bfloat16 is now excluded from the test parameters, hopefully that would make test_core.py to pass. Thank you!

parsifal-47:
Failed twice in pre-commit checks, the reason looks unrelated to the changes, @nhat-nguyen could you please rerun the pipeline, I do not think I have permissions to do that, thank you!

nhat-nguyen (approved):

PR #159, opened 2024-08-06
Implement missing triton operations

parsifal-47:
Tests depend on https://github.com/microsoft/triton-shared/pull/158, once it is merged, I will remove test_precise_math and test_clamp from unsupported list.

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -28,6 +28,7 @@
 #include "llvm/Support/MathExtras.h"
 
 #include <numeric>
+#include <optional>
 #include <type_traits>
 
 using namespace mlir;
@@ -134,10 +135,14 @@ static Value getTransposedValue(Value source, const Location loc,
 }
 
 // for IntLike and FloatLike types
-static unsigned getBitWidth(Type a) {
+static std::optional<unsigned> getBitWidth(Type a) {
   if (auto type = dyn_cast<TensorType>(a))
     return type.getElementType().getIntOrFloatBitWidth();
-  return a.getIntOrFloatBitWidth();
+
+  if (a.isIntOrFloat())
+    return a.getIntOrFloatBitWidth();
+
+  return std::nullopt;
 }
 
 //===----------------------------------------------------------------------===//
@@ -858,7 +863,7 @@ struct FpToFpConverter : public OpConversionPattern<triton::FpToFpOp> {
                   ConversionPatternRewriter &rewriter) const override {
     auto roundingMode = triton::RoundingMode::RTNE; // default
 
-    if (auto roundingModeAttr = op->getAttrOfType<triton::RoundingModeAttr>("rounding")) {
+    if (auto roundingModeAttr = op.getRounding()) {
       roundingMode = roundingModeAttr.getValue();
     }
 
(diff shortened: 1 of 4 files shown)


parsifal-47:
@nhat-nguyen sorry for bothering again, please take a look once you have a chance or let me know whom to ping for reviews. Thank you!

parsifal-47:
Rebased; tests for precise_math and clamp are removed from unsupported list, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -133,6 +133,13 @@ static Value getTransposedValue(Value source, const Location loc,
   return transpose;
 }
 
+// for IntLike and FloatLike types
+static unsigned getBitWidth(Type a) {
```
nhat-nguyen, line 137: should this function return optional in case the type isn't int-like or float-like type to prevent misuse.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -843,6 +850,95 @@ struct BitcastConverter : public OpConversionPattern<triton::BitcastOp> {
...
 
+struct FpToFpConverter : public OpConversionPattern<triton::FpToFpOp> {
+  using OpConversionPattern<triton::FpToFpOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(triton::FpToFpOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto roundingMode = triton::RoundingMode::RTNE; // default
+
+    if (auto roundingModeAttr = op->getAttrOfType<triton::RoundingModeAttr>("rounding")) {
```
nhat-nguyen, line 861: just out of curiosity, is there a way we can get the attribute without having to use the hardcoded string "rounding"?

parsifal-47: thank you for pointing out, I was under impression that this is how it is done because we have multiple instances in this codebase. However, tablegen makes accessor methods and there is no need to lookup strings, fixed now!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: fixed, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -133,6 +134,17 @@ static Value getTransposedValue(Value source, const Location loc,
   return transpose;
 }
 
+// for IntLike and FloatLike types
+static std::optional<unsigned> getBitWidth(Type a) {
+  if (auto type = dyn_cast<TensorType>(a))
+    return type.getElementType().getIntOrFloatBitWidth();
+
```
nhat-nguyen, line 141: i think we're still missing a check if the element type is indeed int or float

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: sweet, much cleaner. thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: right, should be fixed now, thank you!

nhat-nguyen (approved):

PR #164, opened 2024-08-15
Implementing lowering for calls

parsifal-47:
+ removing the assertion which disallowed tensors of 1x1, I do not see why it was set, it seems that removing does not produce an error, let me know what you think. Thank you!

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -868,8 +868,8 @@ struct CallConverter : public OpConversionPattern<triton::CallOp> {
                   ConversionPatternRewriter &rewriter) const override {
     SmallVector<Value> args = adaptor.getOperands();
 
-    // We need to bypass extra arguments added by addProgramInfo which are num_programs and program_ids
-    if (FuncOp parentFunc = op->getParentOfType<FuncOp>()) {
+    // We need to pass extra arguments added by addProgramInfo which are num_programs and program_ids
+    if (FuncOp parentFunc = op->getParentOfType<triton::FuncOp>()) {
       SymbolRefAttr calleeAttr = op.getCalleeAttr();
       StringRef calleeName = calleeAttr.getRootReference();
 
@@ -892,6 +892,11 @@ struct CallConverter : public OpConversionPattern<triton::CallOp> {
 
     auto call = rewriter.create<func::CallOp>(
         op.getLoc(), op.getCallee(), op.getResultTypes(), args);
+
+    if (!call) {
+        op.emitError("Failed to create func::CallOp");
+        return failure();
+    }
 
     rewriter.replaceOp(op, call);
     return success();
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -860,6 +860,44 @@ struct BitcastConverter : public OpConversionPattern<triton::BitcastOp> {
   }
 };
 
+struct CallConverter : public OpConversionPattern<triton::CallOp> {
+  using OpConversionPattern<triton::CallOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(triton::CallOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Value> args = adaptor.getOperands();
+
+    // We need to bypass extra arguments added by addProgramInfo which are num_programs and program_ids
+    if (FuncOp parentFunc = op->getParentOfType<FuncOp>()) {
+      SymbolRefAttr calleeAttr = op.getCalleeAttr();
+      StringRef calleeName = calleeAttr.getRootReference();
+
+      if (ModuleOp module = op->getParentOfType<ModuleOp>()) {
+        if (FuncOp calleeFunc = module.lookupSymbol<FuncOp>(calleeName)) {
+          size_t argsNeed = calleeFunc.getFunctionType().getInputs().size();
+          Block &entryBlock = parentFunc.front();
+          auto parentInputs = entryBlock.getArguments();
+          size_t argsParent = parentInputs.size();
+
+          if (argsNeed > args.size()) {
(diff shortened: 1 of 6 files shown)


parsifal-47:
@nhat-nguyen please take a look when you have a chance, thank you!

nhat-nguyen:
sure, give me some time to double check that removing the assert is ok

parsifal-47:
> sure, give me some time to double check that removing the assert is ok

great, no rush, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -860,6 +860,44 @@ struct BitcastConverter : public OpConversionPattern<triton::BitcastOp> {
...
+  using OpConversionPattern<triton::CallOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(triton::CallOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Value> args = adaptor.getOperands();
+
+    // We need to bypass extra arguments added by addProgramInfo which are num_programs and program_ids
+    if (FuncOp parentFunc = op->getParentOfType<FuncOp>()) {
+      SymbolRefAttr calleeAttr = op.getCalleeAttr();
```
nhat-nguyen, line 873: nit: i know this file has `using namespace triton`, but would you mind prefixing triton::FuncOp to make it clear that we're referring to the triton func op? thank you

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -860,6 +860,44 @@ struct BitcastConverter : public OpConversionPattern<triton::BitcastOp> {
...
+            int missing = argsNeed - args.size();
+            for (int i = 0; i < missing; i++) {
+              args.push_back(parentInputs[parentInputs.size() - i - 1]);
+            }
+          }
+        }
+      }
+    }
+
+    auto call = rewriter.create<func::CallOp>(
```
nhat-nguyen, line 893: should we add error handling here? i would expect each of the if block here to execute every time for valid IRs, but it would be good to handle malformed IRs too

lib/Analysis/MaskAnalysis.cpp
```diff
@@ -315,8 +315,6 @@ LogicalResult MaskState::parseCmp(arith::CmpIOp cmpOp, const Location loc,
       cmpDim = i;
     }
   }
-  assert(cmpDim != -1 &&
```
nhat-nguyen, line 318: Even though removing this assert doesn't trigger any failures, I think the consequence is we end up not applying the mask at runtime. The code below this assert applies the mask to the dimension found in the loop. But since cmpDim == -1, no dimension will match, which is incorrect.

I think we need to understand why the assert is triggered only in the 1x1 case and change the algorithm accordingly. I'd just revert this change and create another PR for this fix since the two changes are not really related.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: I agree, it is better since Functions may belong to different namespaces, fixed, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -860,6 +860,44 @@ struct BitcastConverter : public OpConversionPattern<triton::BitcastOp> {
...
 
+struct CallConverter : public OpConversionPattern<triton::CallOp> {
+  using OpConversionPattern<triton::CallOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(triton::CallOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Value> args = adaptor.getOperands();
+
+    // We need to bypass extra arguments added by addProgramInfo which are num_programs and program_ids
```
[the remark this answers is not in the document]
parsifal-47: fixed, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: added error handling, thank you!

lib/Analysis/MaskAnalysis.cpp
parsifal-47: sounds good, reverted this change, thank you!

parsifal-47:
@nhat-nguyen your comments for this PR should be resolved, thank you for reviewing!

nhat-nguyen (approved):

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -860,6 +860,49 @@ struct BitcastConverter : public OpConversionPattern<triton::BitcastOp> {
...
+        if (FuncOp calleeFunc = module.lookupSymbol<FuncOp>(calleeName)) {
+          size_t argsNeed = calleeFunc.getFunctionType().getInputs().size();
+          Block &entryBlock = parentFunc.front();
+          auto parentInputs = entryBlock.getArguments();
+          size_t argsParent = parentInputs.size();
+
+          if (argsNeed > args.size()) {
+            int missing = argsNeed - args.size();
+            for (int i = 0; i < missing; i++) {
+              args.push_back(parentInputs[parentInputs.size() - i - 1]);
```
JinnDaa, line 886: Hi  parsifal, I wonder why the extra arguments are added **in reverse order** here. Six extra arguments are added by addProgramInfo, and here you add the sixth firstly, and the first finally. Why is the case?

parsifal-47: Hi, thank you for noticing! I do not see any reason for them to be reversed, and I looked at lit test, it clearly shows that they getting backwards, let me fix that quickly

parsifal-47: Thank you @Bolzano983 
https://github.com/microsoft/triton-shared/pull/200

PR #165, opened 2024-08-18
Implementing split, join, and cat

parsifal-47:


parsifal-47:
@nhat-nguyen could you please take a look when you have a chance, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -150,6 +150,36 @@ static std::optional<unsigned> getBitWidth(Type a) {
...
+  for (int i = 0; i < 2; ++i) {
+    offsets.pop_back();
+    sizes.pop_back();
+
+    offsets.push_back(rewriter.getIndexAttr(i));
+    sizes.push_back(rewriter.getIndexAttr(1));
+    op(offsets, sizes, strides, i);
+  }
+}
+
```
nhat-nguyen, line 182: I have a few nit / cosmetic comments about this function:
+ [ ] Should probably rename `processTwoStridesLastDim` to something else. The implementation looks rather fiddly too: are we replacing values of the last 2 dimensions with the first 2 dimensions?
+ [ ] The lambda's arguments should probably have names so that we can better understand what each argument is for.
+ [ ] nit: Our team prefers not to put the size in `SmallVector` declaration
+ [ ] Mixed usage of vector and SmallVector

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -150,6 +150,36 @@ static std::optional<unsigned> getBitWidth(Type a) {
...
+  SmallVector<OpFoldResult, 4> offsets, sizes, strides;
+  std::vector<int64_t> stridesInt(rank, 0);
+
+  int64_t currentStride = 1;
+  for (size_t i = shape.size(); i > 0; --i) {
+    stridesInt[i - 1] = currentStride;
+    currentStride *= stridesInt[i - 1];
+  }
+
```
nhat-nguyen, line 166: Let me know if my understanding is correct: this loop will end up assigning 1s to all elements in `stridesInt` right? `currentStride` starts out at 1, and each step we multiply it again with stridesInt at `i` which is also `1.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -949,6 +979,111 @@ struct PreciseDivConverter : public OpConversionPattern<triton::PreciseDivFOp> {
   }
 };
 
+struct CatConverter : public OpConversionPattern<triton::CatOp> {
```
nhat-nguyen, line 982: i think it would be useful to have a small lit test for each of these new converters so we can understand how each op is being lowered

parsifal-47:
Thank you for the review, will get back with updates!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: thanks for noticing, sorry, refactoring artifact, should be: `currentStride *= shape[i - 1];` so the very last one is 1, and they grow as a product of shape

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: 1. sure, the function assumes there are only two strides for the last dimension and it applies given lambda on these two strides, I have the following options for the name: `operateOnSplitStrides`, `applyToTensorStrides`. Let me know what you think.

Pop-push is there to reuse all data structures between two strided calls, I could evaluate them independently, but that would mean redundant computation for all the dimensions except the last one. What it does now, it computes everything for the whole tensor, then discards the last offset and size, replaces it with 0, 1 for the first stride and 1,1 for the second.
2. done, put names as comments
3. sure, done
4. sorry, fixed

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: sure, will get back with a set of small tests for conversions

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: actually, after some experiments, I think this loop is redundant and all strides are 1, removed

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: added lit tests, but I have mixed feeling about them, it looks like we hardcoding implementation details instead of checking for the behavior

parsifal-47:
all review comments should be addressed, please take a look, thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: You're right that they're hardcoding implementation details, but another benefit is they check 
that the ops are lowered as expected. The lit tests also make it a bit easier to reproduce any issues in the future since the triton IR is already in the mlir format; that means I can just debug issues using the standalone `triton-shared-opt --triton-arith-to-linalg` command instead of having to go through the triton compiler.

test/Conversion/TritonToLinalg/cat.mlir
```diff
@@ -0,0 +1,44 @@
...
+// CHECK:     [[ALLOC_:%.+]] = memref.alloc() : memref<128xf64>
+// CHECK:     memref.copy [[REINTERPRET_CAST_]], [[ALLOC_]] : memref<128xf64, strided<[1]>> to memref<128xf64>
+// CHECK:     [[VAR_0_:%.+]] = bufferization.to_tensor [[ALLOC_]] restrict writable : memref<128xf64>
+// CHECK:     [[REINTERPRET_CAST_0_:%.+]] = memref.reinterpret_cast %arg1 to offset: [0], sizes: [128], strides: [1]{{.*}} : memref<*xf64> to memref<128xf64, strided<[1]>>
+// CHECK:     [[ALLOC_1_:%.+]] = memref.alloc() : memref<128xf64>
+// CHECK:     memref.copy [[REINTERPRET_CAST_0_]], [[ALLOC_1_]] : memref<128xf64, strided<[1]>> to memref<128xf64>
+// CHECK:     [[VAR_1_:%.+]] = bufferization.to_tensor [[ALLOC_1_]] restrict writable : memref<128xf64>
+// CHECK:     [[VAR_2_:%.+]] = tensor.empty() : tensor<256xf64>
+// CHECK:     [[VAR_3_:%.+]] = linalg.generic {indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (-d0 + 127)>], iterator_types = ["parallel"]} ins([[VAR_0_]] : tensor<128xf64>) outs([[VAR_2_]] : tensor<256xf64>) {
+// CHECK:     ^bb0([[IN_:%.+]]: f64, [[OUT_:%.+]]: f64):
```
nhat-nguyen, line 33: i believe the affine map (`-d0 + 127`) for the first value is not right. shouldn't it be just `d0 -> d0`? say we're concatenating lhs (tensor<128>) with rhs (tensor<128>), the first 128 elements should map directly to the first 128 elements in the output; then other 128 elements map to the latter half (d0 + 128)

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: Thank you for resolving the issues. As for `processTwoStridesLastDim`, I still have some questions.

> sure, the function assumes there are only two strides for the last dimension and it applies given lambda on these two strides

The part where the last dimension contains two strides is still not clear to me. Why are there two strides for a single dimension?

> What it does now, it computes everything for the whole tensor, then discards the last offset and size, replaces it with 0, 1 for the first stride and 1,1 for the second.

This is also quite confusing to me. What is the `index` in the lambda supposed to mean? In each loop iteration we're only changing the *last* value of offsets and sizes, and yet the index is `0` and `1`.

I understand the function is trying to reuse code in both `Split` and `Join`, but it's still quite hard for me to grasp exactly what it's trying to encapsulate here.

test/Conversion/TritonToLinalg/cat.mlir
parsifal-47: we cannot do `d0 -> d0` because these two tensors have different number of elements, this would break in runtime, instead I do this trick of 127 - d0 which makes it iterate over the first half of the elements, since the semantics of "cat" does not require to preserver the original ordering, this trick works. 

Summarizing, to concatenate tensor<128> (a) with tensor<128> (b), I need to allocate tensor<256> (c) for results, then copy first half, copy second half. I cannot iterate on a & c with all_to_all affine map because they have different number of elements, I need to cut indexes > 127. Second half is easy to do with d0 + 127, but for the first I do the trick of 127 - d0 which makes it iterate over 0..127. Let me know what you think! Thank you!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: Yes, index here is offset for example, we have a tensor of <128x2xf16>, this lambda runs given function two times with:
(size = (128, 1), offset = (0, 0), stride = (1, 1))
(size = (128, 1), offset = (0, 1), stride = (1, 1))
we always iterate over the big tensor which makes it two strides in this case.

This function I use in one case to copy two small tensors into the big one with extra dimension at the end (which I iterate over). In the second case I copy from this big tensor into the small ones. Since the code for these two operations are mostly the same I put it as a function. Do not hesitate to ask any follow-up questions, I hope that my answers are helping.

test/Conversion/TritonToLinalg/cat.mlir
nhat-nguyen: thank you for the explanation, that makes sense. this makes me a bit uneasy though because we're trying to replicate the triton nvidia compiler's behaviour to make linalg happy.

i spent some time this evening digging around the mlir code base and found that we can actually just use `tensor.concat`. this op also has an integrated rewrite patterns that decomposes to two copies that are more friendly to accelerators (lowers to just a single dma command as opposed to the vectorized loop in the linalg case).

here's the snippet that i used to test:

```mlir
func.func @concat(%arg0: tensor<128xf32>, %arg1 : tensor<128xf32>) -> tensor<256xf32> {
  %1 = tensor.concat dim(0) %arg0, %arg1 : (tensor<128xf32>, tensor<128xf32>) -> tensor<256xf32>
  return %1 : tensor<256xf32>
}

module attributes {transform.with_named_sequence} {
  transform.named_sequence @__transform_main(%root: !transform.any_op {transform.readonly}) {
    %func_op = transform.structured.match ops{["func.func"]} in %root : (!transform.any_op) -> !transform.op<"func.func">
    transform.apply_patterns to %func_op {
      transform.apply_patterns.tensor.decompose_concat
    } : !transform.op<"func.func">
    transform.yield
  }
}
```

the command to lower the `tensor.concat` to two copies:

```
./bin/mlir-opt concat.mlir -transform-interpreter --one-shot-bufferize=allow-return-allocs-from-loops=true
```

parsifal-47: Interesting! I initially started with tensor.concat, but it wasn't properly bufferized. I was not aware about transformational patterns, good to know! Let me get back to you with the code, I am totally fine to rewrite it, thanks a lot for your help!

parsifal-47: done, rewritten with built-in transformational patterns, thanks for the hint!

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -150,6 +150,33 @@ static std::optional<unsigned> getBitWidth(Type a) {
...
+  for (int i = 0; i < 2; ++i) {
+    offsets.pop_back();
+    sizes.pop_back();
+
+    offsets.push_back(rewriter.getIndexAttr(i));
+    sizes.push_back(rewriter.getIndexAttr(1));
+    op(offsets, sizes, strides, i);
+  }
+}
+
```
nhat-nguyen, line 179: I still have some reservations about this function. I understand that it's being used to share code between split and join, but the purpose seems very limited and doesn't offer reuse elsewhere.

At least for me, it makes reading the implementation for split & join much harder since I have to go back and forth between the call site and this implementation. I think the code that we save by having this function is not that much: only the last loop seems non-trivial. The first loop can actually just be simplified by initializing the vector using constructor (except for the sizes). For instance:

```cpp
SmallVector<OpFoldResult> offsets(shape.size(), rewriter.getIndexAttr(0));
SmallVector<OpFoldResult> strides(shape.size(), rewriter.getIndexAttr(1));
SmallVector<OpFoldResult> sizes = llvm::map(...)
```

I think with the above reasons, it might make the code more readable by placing the code directly in the Converter. Let me know what you think.

test/Conversion/TritonToLinalg/join.mlir
nhat-nguyen, line 1: we're going to retire the TritonToLinalg pass soon. Can you move this lit test to TritonArithToLinalg?

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: no problem, updated and substituted, thank you!

parsifal-47:
@nhat-nguyen all your feedback for this PR should be addressed now, thank you!

nhat-nguyen (dismissed):
thank you! 😄

parsifal-47:
> thank you! 😄

Interesting, a lot of unrelated LIT tests are broken because whenever I call `applyTensorConcatDecomposition`, it probably runs canonicalization and removes some of the dead values, for example assert test expects:

```
module {
  func.func @assert_lol(%arg0: i32, %arg1: i32, %arg2: i32, %arg3: i32, %arg4: i32, %arg5: i32, %arg6: i32) {
    %c0_i32 = arith.constant 0 : i32
    %0 = arith.cmpi sgt, %arg0, %c0_i32 : i32
    %1 = tensor.empty() : tensor<1xi1>
    %2 = linalg.fill ins(%0 : i1) outs(%1 : tensor<1xi1>) -> tensor<1xi1>
    cf.assert %0, ".py:0:  Assertion `lol` failed"
    return
  }
```

but it becomes:

```
module {
  func.func @assert_lol(%arg0: i32, %arg1: i32, %arg2: i32, %arg3: i32, %arg4: i32, %arg5: i32, %arg6: i32) {
    %c0_i32 = arith.constant 0 : i32
    %0 = arith.cmpi sgt, %arg0, %c0_i32 : i32
    cf.assert %0, ".py:0:  Assertion `lol` failed"
    return
  }
}
```
LIT test checks for %1 tensor to be there. But I am not sure we need to check for dead values to be present.

What do you think @nhat-nguyen, is it ok for me to drop checks for values that disappeared now from the tests below?
```
Failed Tests (17):
  TRITON-SHARED :: Conversion/TritonArithToLinalg/block_ptr_advance.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/convert_addi_reduce.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/convert_argmin_argmax.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/convert_argmin_argmax_2d.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/convert_minmax_fp_reduce.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/convert_minmax_reduce.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/convert_tensor_reshape.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/dot.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reduce_extend_fp32_precision.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducemax_32_256_bf16.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducesum_512_256_bf16_axis0.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducesum_512_256_bf16_axis1.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducesum_512_256_f32_axis0.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducesum_512_256_f32_axis1.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducesum_middle_dim.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/reducesum_scalar.mlir
  TRITON-SHARED :: Conversion/TritonArithToLinalg/triton_assert.mlir
```

Thank you!

nhat-nguyen:
i think fixing the lit tests makes sense. the dead values are removed from calling "applyPatternsAndFoldGreedily".

parsifal-47:
> i think fixing the lit tests makes sense. the dead values are removed from calling "applyPatternsAndFoldGreedily".

right, I have fixed the tests, it was only a few cases of dead values, in most of the cases it was constant propagation, in some cases tests relied on exact names of the variables. Should be passing now, thank you!

nhat-nguyen (dismissed):

parsifal-47:
rebased to account for the most recent changes

nhat-nguyen (dismissed):

nhat-nguyen:
can you please revert the change that updates the triton submodule?

parsifal-47:
> can you please revert the change that updates the triton submodule? it's what causing the break.

sorry about that, thank you for noticing, reverted!

nhat-nguyen (approved):

PR #170, opened 2024-08-25
Fix tt.dot to support integers

parsifal-47:
Core tests are passing with current triton code, but this change:
https://github.com/triton-lang/triton/pull/4573
should uncover more testcases, I checked that updates tests are also passing.

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -150,6 +150,34 @@ static std::optional<unsigned> getBitWidth(Type a) {
   return std::nullopt;
 }
 
+// true means tensor elements are zeros
+// false means not zero or it cannot be determined
+static bool isZeroTensor(Value &v, bool integers) {
+    if (auto splatOp = v.getDefiningOp<triton::SplatOp>()) {
+      if (auto constOp = splatOp.getSrc().getDefiningOp<arith::ConstantOp>()) {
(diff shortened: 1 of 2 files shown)


include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -150,6 +150,34 @@ static std::optional<unsigned> getBitWidth(Type a) {
...
+          if (integers)
+            return denseAttr.getSplatValue<APInt>().isZero();
+          return denseAttr.getSplatValue<APFloat>().isZero();
+        }
+      }
+    }
+
+    return false;
+}
+
```
nhat-nguyen, line 180: nit: probably cleaner to put this function inside the matmul converter since it's the only user of this function for now.

nhat-nguyen (approved):

parsifal-47:
I have noticed the failure in lit tests, will follow-up with the fix

parsifal-47:
Lit test should be good now

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
parsifal-47: thank you, fixed!

parsifal-47:
I saw: "AttributeError: 'torch.dtype' object has no attribute 'startswith'", but it works for me, looks like I have a different version of pytorch, updated the code

parsifal-47:
failed again on bfloa16, should be fixed now

nhat-nguyen (approved):

PR #171, opened 2024-08-30
Make bitcast work for tensors of pointers

parsifal-47:
I could be missing some high-level ideas about Triton Structured, here is my reasoning:
When bitcasting tensor<ptr<int>> -> tensor<ptr<something_else>> arith::bitcast does not work and all memref/tensor casts. However no manipulations to the data are needed, this is a transformation like casting between different pointer types. I introduced ptr_cast on tts level which becomes unrealized conversion once all types for pointers are settled. This change fixes all comparison tests that require pointer casting at store. Let me know what you think!

### DIFF
--- a/include/triton-shared/AnalysisStructured/PtrAnalysis.h
+++ b/include/triton-shared/AnalysisStructured/PtrAnalysis.h
@@ -242,6 +242,8 @@ public:
   // PtrState for knownPtrs.
   LogicalResult rewriteAddptrOp(triton::AddPtrOp op);
 
+  LogicalResult rewriteBitcastOp(triton::BitcastOp op);
+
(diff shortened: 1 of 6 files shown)


parsifal-47:
among currently open, this is probably the most interesting PR for me in terms of understanding some fundamental ideas @nhat-nguyen could you please take a look when you have a chance

nhat-nguyen:
@haishanzzzz and his team at Meta has an unpublished PR that also handles bitcast, so let's wait until he's back so we can coordinate how to handle this.

parsifal-47:
@haishanzzzz are you back? could you please take a look?

parsifal-47:
@haishanzzzz please take a look when you have a chance

parsifal-47:

PR #180, opened 2024-10-04
Support tensor of indices as loop iter-arg in sequences of pointer arithmetic

nhat-nguyen:
This PR adds support for tensor of indices that are updated in each loop iteration while also being used in pointer arithmetic sequences.

## Approach

Similarly to the pointer types, in the PtrAnalysis pre-pass, we prematurely generate the `tts.get_structured_state` ops for tensor of integers. The important note here is we do not need to know whether these ops will eventually be used in a pointer arithmetic sequence. Any values that are not used in a pointer arithmetic sequence will be removed later in the process.

This approach can easily be extended to other kinds of values that might be used in pointer arithmetic sequences. At a high level, `tts.get_structured_state` can always be used to "wrap" a triton value. This op returns two kinds of values: the first value is always of the same type as the wrapped value, while the remaining values expose the important fields in `PtrState` that are necessary for codegen in scf.for.

The first return value of `tts.get_structured_state` is always an SSA value of the same type as the original value; users of the original triton value will then use this first return value from `tts.get_structured_state` instead. With this approach, even if the original triton value ends up not being used in a pointer arithmetic sequence, it is very easy to revert the IR to the original form by simply deleting the `tts.get_structured_state` op and forwarding the original triton value to its users again.

The other return values then expose the important fields in PtrState that are necessary to generate the code in loops (offsets and strides). Within a loop, for every wrapped triton value returned by a `tts.get_structured_state` op at index `i`, we can always get the corresponding offsets in each loop iteration at index `i + 1` and strides at index `i + 2`.

## Changes

+ Updated the pre-pass to insert `tts.get_structured_state` ops that wrap tensor of indices
+ With the introduction of tensor of indices in loops, we now have to manually visit the `tts.get_structured_state` ops to generate the ops for updating PtrState. We previously did not have to do this because triton pointers always have a `tt.addptr` at the end of each loop, right before yielding the values, which always triggers the process for generating the state-update ops
+ Logic for determining whether a loop iter-arg should have its PtrState updated is improved. We do a BFS-like scan starting from the return values of `tts.get_structured_state` ops to determine if an iter-arg originates from a value that may need its PtrState populated
+ Preliminary support for mask sequences being updated in a loop; this is a bit of a hack and will need more robust implementation if these use cases appear more frequently.
+ Add tests for various scenarios

### DIFF
--- a/lib/AnalysisStructured/PtrAnalysis.cpp
+++ b/lib/AnalysisStructured/PtrAnalysis.cpp
@@ -941,10 +941,8 @@ FailureOr<PtrState> PtrAnalysis::getLoopIterArgPtrState(scf::ForOp forOp,
   }
 
   return reconcileLoopPtrState(
-      forOp, index, state.value(), [](scf::ForOp op, size_t index) {
-        llvm::dbgs() << "getting iter arg index " << index << "\n";
-        return op.getRegionIterArg(index);
-      });
+      forOp, index, state.value(),
+      [](scf::ForOp op, size_t index) { return op.getRegionIterArg(index); });
 }
 
 FailureOr<PtrState> PtrAnalysis::getLoopResultPtrState(scf::ForOp forOp,
--- a/lib/AnalysisStructured/PtrAnalysis.cpp
+++ b/lib/AnalysisStructured/PtrAnalysis.cpp
@@ -1024,38 +1024,31 @@ LogicalResult PtrAnalysis::rewriteForOp(scf::ForOp op) {
 
 LogicalResult
 PtrAnalysis::rewriteGetStructuredStateOp(tts::GetStructuredStateOp op) {
-  auto tritonPtr = op->getOperand(0);
+  auto tritonValue = op->getOperand(0);
 
-  std::optional<tts::PtrState> state;
-  SmallVector<Value> replacements;
-
-  if (knownPtrs.contains(tritonPtr)) {
-    state = knownPtrs[tritonPtr];
-
-    if (ptrMap.contains(tritonPtr)) {
-      Value remappedPtr = ptrMap.lookup(tritonPtr);
-      replacements.push_back(remappedPtr);
-    } else {
-      replacements.push_back(tritonPtr);
-    }
-
-  } else {
-    // If this pointer isn't known, it means PtrAnalysis has failed to analyze
-    // this pointer. In such cases, simply remap all uses of the
-    // structured-pointer back to its original pointer.
+  // If this triton value isn't known, it means PtrAnalysis has failed to
+  // analyze this pointer. In such cases, simply remap all uses of the
+  // structured value back to its original triton value.
+  if (!knownPtrs.contains(tritonValue)) {
     op.emitRemark(
         "Rewrite GetStructuredStateOp failed. Could not find PtrState.");
-    op.getResult(0).replaceAllUsesWith(tritonPtr);
+    op.getResult(0).replaceAllUsesWith(tritonValue);
     return failure();
   }
 
-  if (state->getRank() == 0) {
+  tts::PtrState state = knownPtrs[tritonValue];
+  Value remappedValue =
+      ptrMap.contains(tritonValue) ? ptrMap.lookup(tritonValue) : tritonValue;
+
+  SmallVector<Value> replacements{remappedValue};
+
+  if (state.getRank() == 0) {
     // For scalar pointers, the scalar contains the offset and is the only
     // relevant state that could be updated by the loop.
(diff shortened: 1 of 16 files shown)


nhat-nguyen:
@haishanzzzz Here's the PR for supporting tensor of indices in loop iter-args that are used in pointer sequence. Would appreciate if you could have a look! 😄

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
```diff
@@ -137,21 +136,24 @@ GetStructuredStateOp::getOffsetAndStrideTypes(MLIRContext *context,
...
+GetStructuredStateOp::getOffsetAndStrideSegmentSizes(Type type) {
   int32_t offsetSegmentSize = 0;
   int32_t strideSegmentSize = 0;
 
-  // Unstructured pointers (tensor<!tt.ptr<type>>)
-  if (auto tensorType = llvm::dyn_cast<RankedTensorType>(ptrLikeType)) {
-    if (auto ptrType =
-            dyn_cast<triton::PointerType>(tensorType.getElementType())) {
+  if (auto tensorType = llvm::dyn_cast<RankedTensorType>(type)) {
+    if (tensorType.getElementType().isIntOrIndex()) {
```
robbie-vanderzee-microsoft, line 144: Ditto here, what causes us to only care about `IntOrIndex` types?

robbie-vanderzee-microsoft (dismissed):

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -1131,6 +1146,86 @@ LogicalResult PtrAnalysis::rewriteLoadOp(triton::LoadOp op) {
...
+// detect all such cases, we perform a BFS-like traversal of the IR where the
+// sources are the results of `tts.get_structured_state`. All values that
+// originate from the results of `tts.get_structured_state` are consider
+// "maybeStructured". If a loop's iter-arg is considered "maybeStructured", we
+// must set up their PtrState during `rewriteForOp`.
+void PtrAnalysis::initializeMaybeStructuredArgs(Operation *op) {
+  std::queue<Value> q;
+  DenseSet<Value> visited;
+
+  op->walk([&q, &visited](tts::GetStructuredStateOp op) {
```
haishanzzzz, line 1177: nit, we are overwriting the name `op`

haishanzzzz (approved):
PR LGTM. Thank you Nhat!

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -286,9 +302,11 @@ class TritonToStructuredPass
   }
 
   void runOnOperation() override {
-    if (failed(runTritonToStructuredPrepass())) {
-      signalPassFailure();
-      return;
+    if (!skipPrepass) {
```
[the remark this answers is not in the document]
nhat-nguyen: yup will update in a separate PR with some more comments (i just want to get this in first 😄)

lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -84,7 +85,8 @@ class TritonToStructuredPass
     converter.addConversion([context](RankedTensorType tensorType,
                                       SmallVectorImpl<Type> &types)
                                 -> std::optional<LogicalResult> {
-      if (!isa<triton::PointerType>(tensorType.getElementType())) {
+      if (!isa<triton::PointerType>(tensorType.getElementType()) &&
+          !tensorType.getElementType().isIntOrIndex()) {
```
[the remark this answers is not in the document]
nhat-nguyen: this is because only tensor of int or index can potentially be used to compute pointer offsets; i will add some more comments in a separate PR (just want to get this in first)

lib/Dialect/TritonStructured/IR/TritonStructuredOps.cpp
nhat-nguyen: similarly to the other comment, only tensor of int or index can potentially be used to compute pointer offsets and hence we would only need to use the `tts.get_structured_state` ops for values of such types.

nhat-nguyen:
@robbie-vanderzee-microsoft I just pushed another commit that addresses all your comments. Would you mind resetting your review and giving another approval? Thanks!

robbie-vanderzee-microsoft (commented):
Looks good. Thanks for clarifying!

robbie-vanderzee-microsoft (approved):

PR #181, opened 2024-10-08
Add preliminary support for sign-extend ops in `PtrAnalysis` and `MaskAnalysis`

nhat-nguyen:
This PR adds limited support for sign-extend ops in `PtrAnalysis` and `MaskAnalysis`.

For simplicity, we assume results of the pointer arithmetic operations do not overflow the current platform's native bit-width. This assumption is necessary because most of the memref and tensor related operands use the `index` type whose width, according to the mlir documentation, is platform specific.

Then, we simply need to compute the pointer and mask states using the value whose sign is being extended (i.e.: the visitor of the sign-extend op is just simply forwarding the input operand to its appropriate visitor).

We will revisit the possibility of overflow in the future.

### DIFF
--- a/include/triton-shared/Analysis/MaskAnalysis.h
+++ b/include/triton-shared/Analysis/MaskAnalysis.h
@@ -12,6 +12,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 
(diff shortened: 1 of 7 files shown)


robbie-vanderzee-microsoft (approved):
Looks good.

[!NOTE]
#180 should probably be accepted after.

PR #182, opened 2024-10-08
Support triton kernels containing early returns

nhat-nguyen:
Previously, the `triton-arith-to-linalg` pass always converts terminator ops to `func.return`. With the presence of early returns, `cf.br` ops can also appear in the IR; these `cf.br` ops are also considered terminators and are accidentally converted to `func.return`. I have modified the lowering logic to only convert `tt.return` ops.

A simple CPU test case is also added. As part of this test, I have discovered that `eliminate-empty-tensors` does not support functions with multiple return ops. Since this is mostly an optimization pass to reduce the number of allocations, I have disabled the pass for now. We will revisit this again if the need arises.

### DIFF
--- a/backend/compiler.py
+++ b/backend/compiler.py
@@ -52,7 +52,12 @@ def _ttsharedir_to_llir(ttsharedir: str):
         # TritonShared-MLIR to LLVM-MLIR
         subprocess.check_call([mlir_opt_path, ttshared_path,
             "--convert-linalg-to-affine-loops",
-            "--eliminate-empty-tensors",
+            # Note: eliminate-empty-tensors fails when there are multiple func.return ops
+            # in a single kernel which are the results of early returns.
+            # See python/examples/test_early_return.py for examples.
+            # We disable this pass for now since performance on CPU isn't the main
+            # focus at the moment.
+            # "--eliminate-empty-tensors",
             "--empty-tensor-to-alloc-tensor",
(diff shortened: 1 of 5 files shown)


beicy (dismissed):

backend/compiler.py
```diff
@@ -52,7 +52,12 @@ def _ttsharedir_to_llir(ttsharedir: str):
         # TritonShared-MLIR to LLVM-MLIR
         subprocess.check_call([mlir_opt_path, ttshared_path,
             "--convert-linalg-to-affine-loops",
-            "--eliminate-empty-tensors",
+            # Note: eliminate-empty-tensors fails when there are multiple func.return ops
```
robbie-vanderzee-microsoft, line 55: Maybe add todo?

python/examples/test_early_return.py
```diff
@@ -0,0 +1,55 @@
+import torch
+
+import triton
+import triton.language as tl
+
+from triton.backends.triton_shared.driver import CPUDriver
+
+@triton.jit
+def early_return(in0, out0):
```
robbie-vanderzee-microsoft, line 9: Maybe add a test which has multiple conditional returns?

robbie-vanderzee-microsoft (approved):
Looks good.

beicy (approved):

PR #183, opened 2024-10-09
Lower cumsum on integer types

nhat-nguyen:
We previously only pattern match `tt.scan` op containing an `arith.addf`. Add support for `arith.addi` as well.

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -1741,6 +1741,10 @@ struct DenseConstantConverter : public OpConversionPattern<arith::ConstantOp> {
 class CumSumConverter : public OpConversionPattern<triton::ScanOp> {
   using OpConversionPattern<triton::ScanOp>::OpConversionPattern;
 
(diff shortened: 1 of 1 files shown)


include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -1741,6 +1741,10 @@ struct DenseConstantConverter : public OpConversionPattern<arith::ConstantOp> {
 class CumSumConverter : public OpConversionPattern<triton::ScanOp> {
   using OpConversionPattern<triton::ScanOp>::OpConversionPattern;
 
+  bool isAddOp(Operation *op) const {
```
CaptnJackSparrow, line 1744: I would just remove this function and use `isa` directly in the code below (better readability) especially since it is quite simple.

CaptnJackSparrow (approved):

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: thanks! i simplified the code, looks much better now

beicy (approved):

PR #186, opened 2024-10-18
Fix mask analysis for when the entire tensor is masked off

nhat-nguyen:
The current formula for computing masks does not work when the mask bound is smaller than the start of the mask range:

```
---|-------|-----------|
   ^       ^           ^
bound    start        end
```

Current formula:

```
new_end = min(end, bound)
new_dim = new_end - start
```

For the above case, this formula will produce a negative `new_dim`. To fix this issue, we optionally move `new_end` back to `start` so that when `bound < start`, `new_dim` is 0.

The new formula is:

```
new_end_tmp = min(end, bound)
new_end = max(new_end_tmp, start)
new_dim = new_end - start
```

Another formula that could work in theory is to do:

```
new_end = min(end, bound)
new_dim_potentially_neg = new_end - start
new_dim = max(new_dim_potentially_neg, 0)
```

But this approach does not work in MaskAnalysis because we operate on the `index` type which is unsigned. We would have a negative overflow when computing `new_dim_potentially_neg` and end up getting a positive number instead.

# Changes
+ Update the formula
  + The change is quite invasive, so I added a flag in cases we don't want to enable this fix
+ Update lit tests
  + Removed some of the old TritonToLinalg tests; we will remove the old pass in a future PR


beicy (approved):

Busterkinng88:
🤴

PR #189, opened 2024-11-05
Set PtrState scalar to 0 for pointers from kernel arguments

nhat-nguyen:
When rewriting `tts.get_structured_state` operating on scalar pointers, we need to use the offset from `state.scalar`. Pointers directly from the kernel arguments unfortunately is missing the `scalar` field, causing a crash when creating offset from this `scalar` field.


beicy (approved):
LGTM!

PR #202, opened 2024-12-19
Support mask analysis against 0, and between two scalars

kile01:
Support was added for two new scenarios:
`arith.cmpi ge %scalar, %c0`: aka offset comparison to the lower bound of 0. Mask analysis already has an implicit assumption that the beginning of a mask starts at 0, so support was added to allow this case through and assumes that this comparison evaluates to true.
`arith.cmpi slt %scalar,_1 %scalar_2`: offset comparison between two scalars. E.g.:
```
%11 = tt.expand_dims %offset
%cst_4 = arith.constant dense<324> : tensor<16x1xi64>
%23 = arith.cmpi slt, %11, %cst_4 : tensor<16x1xi64>
```
This example is notable in that we cannot take the normal approach of computing the minimum of the lhs and rhs as the new dimension (the lhs offset may be 0). To handle this, a ternary operator is inserted to evaluate the comparison at runtime. If it succeeds, we keep the existing dimensions from the lhs, otherwise we assume nothing should be loaded/stored.

This change also adds a dump method to both `MaskState` and `PtrState` as a small QOL improvement.

### DIFF
--- a/include/triton-shared/Analysis/MaskAnalysis.h
+++ b/include/triton-shared/Analysis/MaskAnalysis.h
@@ -120,14 +120,14 @@ private:
                          OpBuilder &builder);
 
   // Operand is the result of cmpi
-  // Assume only one of the dimensions has size > 1. Only support slt and sge
-  // against 0 for now. For that dimension, we have three cases:
+  // Assume only one of the dimensions has size > 1. Only support slt/ult, and
+  // sge against 0 for now. For that dimension, we have three cases:
   //  1. Constant comparison with both left and right-hand sides being scalars.
   //     Calculate this new dim as a compare and select.
   //      I.e. dim = lhs < rhs ? end : 0
   //  2. Left-hand side is not a scalar, and the right-hand side is.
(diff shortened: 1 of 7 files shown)


kile01:

nhat-nguyen:
@kile01 Thank you Kirsten! Sorry for the delay as I went on vacation. Looks like there's a typo in the lit test command. The command we use is:

```
RUN: triton-shared-opt ... %s | FileCheck %s
```

kile01:
> triton-shared-opt

No worries, this isn't urgent :) and thank you for the catch!

nhat-nguyen (dismissed):
Changes look good to me! This is optional, but would be great if we could also have a small snippet of the triton program that generates the IR (I usually put this as a comment in the lit test)

kile01:
thanks for the review @nhat-nguyen! I've added the example program as a comment

kile01:
@nhat-nguyen would you mind doing the honors and merging the PR? I don't seem to have write access

nhat-nguyen (approved):

PR #205, opened 2024-12-29
Change the ordering of `cse` and `canonicalize` passes for `cse` to be more efficient

parsifal-47:
--cse before --canonicalize does not look optimal, but this is the original sequence from triton-to-linalg-experimental

### DIFF
--- a/backend/compiler.py
+++ b/backend/compiler.py
@@ -33,12 +33,7 @@ def _ttir_to_ttsharedir(mod):
         dst_path = os.path.join(tmpdir, "ttshared.mlir")
         Path(src_path).write_text(ttir_code)
         triton_shared_opt_path = _get_triton_shared_opt_path()
-        subprocess.check_call([triton_shared_opt_path, src_path,
-            "--triton-to-structured",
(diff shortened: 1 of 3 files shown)


nhat-nguyen:
thank you! could we swap this ordering in triton-to-linalg-experimental instead?

parsifal-47:
> thank you! could we swap this ordering in triton-to-linalg-experimental instead?

Sure, updated, please take a look, thank you!

parsifal-47:
I have updated LIT test to match new IR structure caused by different pass ordering

nhat-nguyen:
just out of curiosity, do you know what the IR would look like if we run another canonicalization after cse? also we should update the PR title to reflect the new changes

parsifal-47:
> just out of curiosity, do you know what the IR would look like if we run another canonicalization after cse? also we should update the PR title to reflect the new changes

That is a good question, I tried adding one more canonicalization and it did not change anything for our LIT test, so likely results are not very dramatic. PR title updated, thank you!

parsifal-47:

PR #210, opened 2025-01-02
Introduce `triton-to-unstructured` pass

nhat-nguyen:
This PR introduces the `triton-to-unstructured` pass which is the first step towards allowing triton-shared to compile pointer sequences that cannot be analyzed by `triton-to-structured` (gather / scatter).

This pass attempts to lower all loads and stores of unstructured pointers to
tts.gather or tts.scatter that take a single base, a tensor of offsets, an
optional tensor of mask values, and a default value in case of load.

In addition, all pointer-producing ops will be eliminated and replaced by
offset-producing ops. tts.gather and tts.scatter will use the pointer
directly from the kernel arguments as opposed to pointer produced by ops such
as tt.addptr and tt.splat.

Example:

```mlir
module {
  tt.func public @gather_simple_no_loop(%arg0: !tt.ptr<f32>, %arg1: !tt.ptr<f32>) attributes {noinline = false} {
    %cst = arith.constant dense<5> : tensor<64xi32>
    %cst_0 = arith.constant dense<10> : tensor<64xi32>
    %0 = tt.make_range {end = 64 : i32, start = 0 : i32} : tensor<64xi32>
    %1 = arith.divsi %0, %cst_0 : tensor<64xi32>
    %2 = arith.addi %1, %cst : tensor<64xi32>
    %3 = tt.splat %arg0 : !tt.ptr<f32> -> tensor<64x!tt.ptr<f32>>
    %4 = tt.addptr %3, %2 : tensor<64x!tt.ptr<f32>>, tensor<64xi32>
    %5 = tt.load %4 : tensor<64x!tt.ptr<f32>>
    %6 = tt.splat %arg1 : !tt.ptr<f32> -> tensor<64x!tt.ptr<f32>>
    %7 = tt.addptr %6, %0 : tensor<64x!tt.ptr<f32>>, tensor<64xi32>
    tt.store %7, %5 : tensor<64x!tt.ptr<f32>>
    tt.return
  }
}
```

becomes

```mlir
module {
  tt.func public @gather_simple_no_loop(%arg0: !tt.ptr<f32>, %arg1: !tt.ptr<f32>) attributes {noinline = false} {
    %cst = arith.constant dense<5> : tensor<64xi32>
    %cst_0 = arith.constant dense<10> : tensor<64xi32>
    %0 = tt.make_range {end = 64 : i32, start = 0 : i32} : tensor<64xi32>
    %1 = arith.divsi %0, %cst_0 : tensor<64xi32>
    %2 = arith.addi %1, %cst : tensor<64xi32>
    %3 = tts.gather %arg0[%2] : (<f32>, tensor<64xi32>) -> tensor<64xf32>
    tts.scatter %3 into %arg1[%0] : tensor<64xf32> into (<f32>, tensor<64xi32>)
    tt.return
  }
}
```

Current assumptions and limitations:

- For simplicity, the pass assumes that gather / scatter operations load /
store from / to a single base with a tensor of random offsets. As a
result, the following triton program would not work:

```python
@triton.jit
def gather_simple(in0, in1, out0):
    offs = tl.arange(0, 8)
    in0_ptrs = in0 + offs
    in1_ptrs = in1 + offs
    ptrs = tl.cat(in0_ptrs, in1_ptrs, can_reorder=True)
    c = tl.load(ptrs)
    out_offs = tl.arange(0, 16)
    tl.store(out0 + out_offs, c)
```

In the above program, `ptrs` contains 2 bases: `in0` and `in1` after the
`cat` operation.

For more details on the algorithm, see the `TritonToUnstructuredPass.cpp` file.

# Future work

Future work may include scaling the algorithm to support multiple bases -- one
possible solution is to let tts.gather and tts.scatter take in an additional
tensor of base pointers corresponding to the tensor of offsets. But because
we do not want pointer-producing ops to be present after this pass, we can
use a tensor of index where each element indicates the index of the pointer
argument to be used. The drawback is a gather or scatter operation now needs
one extract lookup to get the base which will affect performance.

---

# Intended lowering pipeline
- triton-to-structured (no changes):
    - analyzes structured addptr sequences
        - introduces `tts.make_tptr %ptr_arg with offsets and strides`
        - introduces `tts.load` and `tts.store`
    - leaves unstructured addptr sequences and their corresponding `tt.load` and `tt.store` intact
- triton-to-unstructured (#210):
    - introduces `tts.gather` and `tts.scatter`
    - removes all pointer-producing ops such as `tt.addptr` and `tt.splat` and replaces them with offset-producing ops
- structured-to-memref (#217):
    - currently converts everything to memref including scalar addptr and kernel arguments
    - will change to just convert ops in the `tts` dialect to `memref` with the exception of `tts.gather` and `tts.scatter`
- unstructured-to-memref (#216):
    - converts the remaining unstructured `tts.gather`, `tts.scatter` into memref
- triton-ptr-to-memref (#211):
    - converts kernel arguments with pointer type to memref


include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -145,6 +143,57 @@ def TTS_GetStructuredStateOp : TTS_Op<"get_structured_state", [AttrSizedResultSe
...
+    attr-dict `:` `(` type($ptr) `,` type($offset) `)` `->` type($result)
+  }];
+}
+
+def TTS_ScatterOp : TTS_Op<"scatter", [
+  MemoryEffects<[MemWrite]>,
+  OptionalTypesMatchWith<"mask type matches offset type", "offset", "mask",
+                 "triton::getI1SameShape($_self)">
+]> {
+  let summary = "optionally load data from in memory to fill a portion of the tensor";
```
python3kgae, line 181: store instead of load?

nhat-nguyen: thanks! fixed :)

haishanzzzz:
Thank you for the PR! I have a few general questions:

1) The transformation this pass performs seems to be targeting hardware that uses "base ptr + a vector of offset" in performing gather/scatter, while GPU-like architectures simply create a tensor of addresses. Is that correct?

2) Would there still be kernels that can't be lowered to this pattern, even though very unlikely? E.g., would it be possible to have a vector of int64s from memory, convert them to pointers, and use it to perform ld/st?

3) You mentioned one future work is "scaling the algorithm to support multiple bases". I thought it is a very promising direction. Do you have specific ideas / timeline in mind? Would love to discuss more offline.

kile01:
I may have missed it, but do you have any examples of the `tts.gather` with a default value?

lib/Conversion/TritonToUnstructured/TritonToUnstructuredPass.cpp
```diff
@@ -0,0 +1,544 @@
...
+                  ptrUsers.push_back(op);
+                  return success();
+                })
+                .Case<scf::ForOp>([&](scf::ForOp forOp) {
+                  // Index of the init-arg corresponding to this use, note that
+                  // we have to subtract by 3 from the operand number because
+                  // scf.for ops always have 3 leading operands for start, end,
+                  // and step.
+                  auto argIndex = use.getOperandNumber() - 3;
+                  auto init = forOp.getInitArgs()[argIndex];
```
kile01, line 374: do you need an early exit if there are no init args present?

nhat-nguyen:
> 3\. You mentioned one future work is "scaling the algorithm to support multiple bases". I thought it is a very promising direction. Do you have specific ideas / timeline in mind? Would love to discuss more offline.

Oh I do have these but in another PR, let me copy them over to this PR too. Thanks for noticing :D

kile01 (approved):
LGTM!

lib/Conversion/TritonToUnstructured/TritonToUnstructuredPass.cpp
```diff
@@ -0,0 +1,544 @@
...
+//   - a tensor of offsets (or single offset) that indicates the offsets from
+//   the base pointer
+//
+// All intermediate tt.addptr ops are converted to arith.addi ops that compute
+// the offsets. Offsets start at 0 with the provided bit-width. All pointer
+// shape manipulation ops such as tt.splat and tt.broadcast will instead operate
+// on the offsets and will be converted to linalg in triton-arith-to-linalg.
+//
+// By default, the pass uses i32 for the initial offsets of all pointers
+// (configurable via offset-bit-width=width). If any intermediate tt.addptr
```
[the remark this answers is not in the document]
nhat-nguyen: I was debating between using the offsets as they are (the current solution) vs. convert the offsets to index and found that the former results in more readable IR. This is because the latter approach always requires inserting arith.index_cast for the offset for each addptr.

include/triton-shared/Conversion/TritonToUnstructured/Passes.h
```diff
@@ -0,0 +1,15 @@
+#ifndef TRITON_TO_UNSTRUCTURED_CONVERSION_PASSES_H
+#define TRITON_TO_UNSTRUCTURED_CONVERSION_PASSES_H
+
+#include "triton-shared/Conversion/TritonToUnstructured/TritonToUnstructured.h"
+
+namespace mlir {
+namespace triton {
+
+#define GEN_PASS_REGISTRATION
```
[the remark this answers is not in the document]
nhat-nguyen: Thank you! It definitely sounds cleaner. I quickly tried but it didn't work -- I'm probably missing something. I might just leave these as they are now and do a one pass clean up in the future because all of the passes here use the same convention anyway. 😄

lib/Conversion/TritonToUnstructured/TritonToUnstructuredPass.cpp
nhat-nguyen: I think if we're in this block, the init args are definitely non-empty because the `for` op is a user of the current value. Hence that value must be in the init-args

beicy (approved):
LGTM!

include/triton-shared/Conversion/TritonToUnstructured/Passes.h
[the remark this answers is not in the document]
kile01: That makes sense 😄 this was something I recently found out about and thought I'd share

PR #211, opened 2025-01-02
Introduce `triton-ptr-to-memref` pass

nhat-nguyen:
This PR introduces the `triton-ptr-to-memref` pass responsible for converting function signature that uses triton ptr to use memref instead. This is part of the work to allow triton-shared to lower gather / scatter pointer sequences.

Much of this code is copied from the current `StructuredToMemref` pass which will be cleaned up in a later PR.

---

# Intended lowering pipeline
- triton-to-structured (no changes):
    - analyzes structured addptr sequences
        - introduces `tts.make_tptr %ptr_arg with offsets and strides`
        - introduces `tts.load` and `tts.store`
    - leaves unstructured addptr sequences and their corresponding `tt.load` and `tt.store` intact
- triton-to-unstructured (#210):
    - introduces `tts.gather` and `tts.scatter`
    - removes all pointer-producing ops such as `tt.addptr` and `tt.splat` and replaces them with offset-producing ops
- structured-to-memref (#217):
    - currently converts everything to memref including scalar addptr and kernel arguments
    - will change to just convert ops in the `tts` dialect to `memref` with the exception of `tts.gather` and `tts.scatter`
- unstructured-to-memref (#216):
    - converts the remaining unstructured `tts.gather`, `tts.scatter` into memref
- triton-ptr-to-memref (#211):
    - converts kernel arguments with pointer type to memref


include/triton-shared/Conversion/TritonPtrToMemref/Passes.td
```diff
@@ -0,0 +1,11 @@
+#ifndef TRITON_PTR_TO_MEMREF_CONVERSION_PASSES
+#define TRITON_PTR_TO_MEMREF_CONVERSION_PASSES
+
+include "mlir/Pass/PassBase.td"
+
+def TritonPtrToMemref : Pass<"triton-ptr-to-memref", "mlir::ModuleOp"> {
+  let summary = "Convert Triton non-block pointer to TritonStructured dialect";
```
red1bluelost, line 7: Looks like this is the summary from TritonToStructured.

lib/Conversion/TritonPtrToMemref/TritonPtrToMemrefPass.cpp
```diff
@@ -0,0 +1,130 @@
...
+    ConversionTarget target(getContext());
+    TritonFunctionSignatureConverter typeConverter;
+
+    // Update function signature and call ops to use memrefs
+    target.addDynamicallyLegalOp<func::FuncOp>([&](func::FuncOp op) {
+      return typeConverter.isSignatureLegal(op.getFunctionType());
+    });
+    target.addDynamicallyLegalOp<triton::FuncOp>([&](triton::FuncOp op) {
+      return typeConverter.isSignatureLegal(op.getFunctionType());
+    });
```
red1bluelost, line 101: (minor) I think you could combine these two by just having two types in the template parameter and changing the function call to accept `auto`. (Unless this API is on a newer version of MLIR than what is currently used here.)

lib/Conversion/TritonPtrToMemref/TritonPtrToMemrefPass.cpp
```diff
@@ -0,0 +1,130 @@
...
+    target.addDynamicallyLegalOp<func::CallOp>([&](func::CallOp op) {
```
red1bluelost, line 102: I don't think I saw any tests that had a call op. Is this something worth testing if so should it be added to this PR or a later PR?

lib/Conversion/TritonPtrToMemref/TritonPtrToMemrefPass.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+namespace {
+
+class TritonFunctionSignatureConverter : public TypeConverter {
+public:
+  TritonFunctionSignatureConverter() {
+    // The order of type conversion is important: later ones are tried earlier.
+    addConversion([](Type type) { return type; });
+    addConversion([](triton::PointerType ptrType) {
+      return UnrankedMemRefType::get(ptrType.getPointeeType(), 0);
```
red1bluelost, line 50: (minor) Could you add a `/*parameterName=*/` style comment since the zero isn't clear why it's used as a parameter here. (like in case 4 of the LLVM coding standards https://llvm.org/docs/CodingStandards.html#comment-formatting)

lib/Conversion/TritonPtrToMemref/TritonPtrToMemrefPass.cpp
```diff
@@ -0,0 +1,130 @@
...
+      return builder.create<UnrealizedConversionCastOp>(loc, resultType, inputs)
+          .getResult(0);
+    });
+
+    addArgumentMaterialization([&](OpBuilder &builder, Type resultType,
+                                   ValueRange inputs,
+                                   Location loc) -> std::optional<Value> {
+      return builder.create<UnrealizedConversionCastOp>(loc, resultType, inputs)
+          .getResult(0);
+    });
```
red1bluelost, line 73: (nit) Could you outline these two lambda's into a local variable since they share the same code.

red1bluelost (approved):
Looks good to me, just some minor comments.

beicy (approved):
LGTM!

beicy (approved):
LGTM!

PR #222, opened 2025-01-23
Fix missing cmake build dependency in `MaskAnalysis`

nhat-nguyen:
The MaskAnalysis file depends on the triton structured dialect, but it isn't specified in the cmake file. This caused build failure in the rare cases.


NathanielMcVicar (approved):
Nice catch!

PR #225, opened 2025-02-04
[Doc] Update build instruction

python3kgae:
Add pybind11 and setuptools to the pip install package list. 
Fix the directory for virtualenv build.


nhat-nguyen (approved):
thank you :)

PR #229, opened 2025-02-19
Remove unused TritonAnalysis and TritonGPU dependencies

Nullkooland:
Modify CMake to remove unused `TritonAnalysis` and `TritonGPU` related depedencies.


nhat-nguyen (approved):
Thanks so much for the cleanup 💯

PR #239, opened 2025-03-21
Support for Gather/Scatter access with only one non-continuous dimension

python3kgae:
A new operation, TTS_MakeGatherScatterTensorPtrOp, was added.
It is similar to TTS_MakeTensorPtrOp, but includes gather_scatter_dim and gather_scatter_offset to store information about the non-continuous dimension.
With TTS_MakeIndirectTensorPtrOp, the continuous memory accesses for the rest of the dimensions are grouped into one operation, which keeps the high-level semantics of the operation intact.

In PtrAnalysis:
  If an operation works on a single dimension, unsupported operations will have PtrState start from the operation with the operation as the indirect offset.
  For other operations that fail the structured check, reset the dimension using the operation as the indirect offset if it works on a single dimension.

When dealing with mulState and addState:
  For the indirect dimension, set the stride to 1 and use offset * stride as the new offset.
  If one state has a modulo and the other state is not structured, clear the modulo and use the operand as the offset directly.

Also skipped loop case when the PtrState is not structured.


bmyerz0:
I think the naming could be clarified for the op and tests. The concepts of a gather/scatter and indirect pointer seem distinct to me. "Indirect" makes me think of pointer that is derived from a load. "Gather/scatter" makes me think of a collection of offsets into a data structure.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -74,8 +74,111 @@ bool PtrState::dimHasModulo(uint32_t dim) const {
...
+  for (int32_t i = 0; i < getRank(); i++) {
+    if (!dimIsIndirect(i))
+      continue;
+    dims.emplace_back(i);
+  }
+  assert(dims.size() == 1 && "must have single indirect dimension");
+  return dims.front();
+}
+
+bool PtrState::noContinuousDim() const {
```
bmyerz0, line 100: what happens if getRank() > 1 and dimsIsIndirect for all dims?

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -74,8 +74,111 @@ bool PtrState::dimHasModulo(uint32_t dim) const {
...
+  auto value = dyn_cast<Value>(offsets[dim]);
+
+  return isa<ShapedType>(value.getType());
+}
+
+int32_t PtrState::getIndirectDim() const {
+  SmallVector<int32_t> dims;
+  for (int32_t i = 0; i < getRank(); i++) {
+    if (!dimIsIndirect(i))
+      continue;
```
bmyerz0, line 93: nit: this looks like it could use `std::copy_if` or something similar

bmyerz0:
Can you explain the decision to set stride to 1? Is it going to be unused? Or is there a technical reason in the analysis? If unused, can we pick a failsafe value?

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -74,8 +74,111 @@ bool PtrState::dimHasModulo(uint32_t dim) const {
   return intAttr.value() != 0;
 }
 
+bool PtrState::dimIsIndirect(uint32_t dim) const {
+  assert(dim < getRank());
+
+  auto intAttr = getIntAttr(offsets[dim]);
+  if (intAttr && intAttr.has_value()) {
```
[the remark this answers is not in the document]

lib/AnalysisStructured/PtrAnalysis.cpp
python3kgae: Want to use copy_if with std::range here.
But std::range is c++ 20 only :(
Do you have suggestion for something else?

lib/AnalysisStructured/PtrAnalysis.cpp
python3kgae: Updated to check all dimensions.

python3kgae:
> I think the naming could be clarified for the op and tests. The concepts of a gather/scatter and indirect pointer seem distinct to me. "Indirect" makes me think of pointer that is derived from a load. "Gather/scatter" makes me think of a collection of offsets into a data structure. This proposal seems closer to gather/scatter.

Updated to gather_scatter.

python3kgae:
> Can you explain the decision to set stride to 1? Is it going to be unused? Or is there an intended reason in the ptr analysis? If unused, can we pick a failsafe value?

In the normal case, such as `ptr = base_ptr + offset * stride`, we identify the offset as `offset` and the stride as `stride`.
However, in cases like `ptr = base_ptr + offset0 * stride0 + offset1 * stride1`, we cannot identify the individual offsets and strides.
Therefore, we take the combined offset as `offset0 * stride0 + offset1 * stride1` and set the stride to `1`. The pointer is still in the form of `base_ptr + offset * stride` (with the offset being `offset0 * stride0 + offset1 * stride1` and the stride being `1`).

If a special failsafe value is set for stride, we need to check this value and then use the offset directly. Instead of setting the stride to 1, there is no need to check the special value.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -120,6 +120,77 @@ def TTS_MakeTensorPtrOp
...
+  //       Not support case where the offset is a tensor load from other ptr which for multiple dimension.
+  //
+  //       offset_m = tl.arange(0, M)
+  //       offset_n = tl.arange(0, N)
+  //       offset_k = tl.arange(0, K)
+  //       ld_offsets = tl.load(a_ptr + offset_m[:,None]+offsets_n[None,:])
+  //       not_support = tl.load(b_ptr + ld_offsets)
+  //       not_support2 = tl.load(b_ptr + ld_offsets * (offset_m[:,None]+offsets_n[None,:]))
+  //       not_support3 = tl.load(b_ptr + (ld_offsets * (offset_m[:,None]+offsets_n[None,:]))[:, :, None] + offset_k[None,None,:])
+
```
nhat-nguyen, line 135: I think "only support case that the offset could be identified for each dimension." is not clear. Isn't it true that we always have the offsets for all dimensions?

It's also worth pointing out the supported cases where one of the dimension is structured while the other is not.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -120,6 +120,77 @@ def TTS_MakeTensorPtrOp
...
+  let summary = "create an pointer that points to a tensor in memory for gather/scatter";
+  let description = [{
+    The `tts.make_gather_scatter_tptr` operation is similar to `tts.make_tptr`.
+    The key difference is that `make_gather_scatter_tptr` accesses the tensor non-continuously.
+    Currently, only one dimension is allowed to be non-continuous.
+    This dimension is saved in `gather_scatter_dim`, and the offset for that dimension is saved in `gather_scatter_offset`.
```
nhat-nguyen, line 141: I think it's also worth pointing out that each contiguous load will load from this offset.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -99,14 +205,47 @@ LogicalResult PtrState::addState(const PtrState &lhsState,
...
+      return failure();
+    }
+  }
 
-    auto newStride =
-        addOFRs(lhsState.strides[i], rhsState.strides[i], loc, builder);
-    strides.push_back(newStride);
+  for (uint64_t i = 0; i < lhsState.getRank(); i++) {
+    if (!lhsState.dimIsNotContinuous(i) && !rhsState.dimIsNotContinuous(i)) {
+      auto newOffset =
```
nhat-nguyen, line 218: nit: I find having negative in function names make it hard to read, specifically for case like these where you end up doing `!notSomething()`. Can you change all of the methods to check for continuous instead?

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -297,6 +482,16 @@ LogicalResult PtrAnalysis::visitOperandAdd(arith::AddIOp addOp, PtrState &state,
     return failure();
   }
 
+  // When one state hasModulo while other state is not structured.
+  // Need to clear the modulo and use the operand as offset directly.
+  if (!lhsState.isStructured() && rhsState.hasModulo()) {
```
nhat-nguyen, line 487: does this mean modulo is not supported? can we support this?

nhat-nguyen: If lhs is not structured and rhs is structured but has modulo, my opinion is we still want to have support. We can do this in a separate PR to reduce complexity.

lib/AnalysisStructured/PtrAnalysis.cpp
bmyerz0: I think using copy_if with dims.begin(), dims.end() would work

bmyerz0: ah, well no worries then; I don't have something else in mind

lib/AnalysisStructured/PtrAnalysis.cpp
python3kgae: Added TODO for this.

include/triton-shared/AnalysisStructured/PtrAnalysis.h
```diff
@@ -57,10 +57,31 @@ struct PtrState {
 
   bool dimHasModulo(uint32_t dim) const;
 
+  bool dimIsNotContinuous(uint32_t dim) const;
+  int32_t getNonContinuousDim() const;
+  // When rank is 1, and the only dimension is not continuous.
+  // There's no dimension is continuous.
+  bool noContinuousDim() const;
```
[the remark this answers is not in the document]
python3kgae: Updated.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
python3kgae: Updated.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
python3kgae: Updated.

nhat-nguyen (approved):
Landing this for now. We will address any other feedback from the community later as necessary

PR #246, opened 2025-04-02
Remove 0-init of accumulation tensor by default

CaptnJackSparrow:


include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -1145,33 +1145,16 @@ struct MatmulConverter : public OpConversionPattern<triton::DotOp> {
...
-    auto zero = rewriter.create<mlir::arith::ConstantOp>(
-        op.getLoc(), elementType, constantAttr);
-
-    auto zeroes =
-        rewriter.create<linalg::FillOp>(loc, ValueRange{zero}, ValueRange{init})
-            .result();
+    if (isZeroTensor(opc, integers)) {
+      opc = rewriter.create<tensor::EmptyOp>(loc, dstType.getShape(),
+                                             elementType);
+    }
```
red1bluelost, line 1151: The contents of `tensor.empty` is unspecified so this would break the behavior if the matmul opc is zeros.

nhat-nguyen: Yeah the fill op will be necessary for testing on the cpu backend; I'd probably make this a pass option and let the caller decide whether to zero-initialize the output.

Another thing I'm worried about is the direct use of opc as the accumulator. The bufferization pass will look at the `outs` (opc) in this case and decide whether to reuse the buffer or not. If this `opc` is used elsewhere, we might have incorrect codegen here. It has been a while since I looked at bufferization so not sure if this is still the case.

CaptnJackSparrow: Here is the justification behind this change, there are 2 cases we need to support:

1. If C == 0: return A . B
2. Else: return A . B + C

In case 1, it does not make sense to have a fill op to set C to 0 as the values in C is not required. This is a waste of compute, so randomly initialized C is perfectly fine, we just need to write the result of the matmul to C.

So really the compiler should just distinguish between the 2 cases by checking whether the opC is an empty tensor or not and the lowering passes should handle this accordingly. Filling C to 0 is not required by either cases. The Maia compiler already does this, and accumulation on the C tensor works accordingly. Does this make sense?

nhat-nguyen: I think the problem here is in maia, we don't lower linalg.matmul into loops. But for other users, linalg.matmul will be lowered to affine which accumulates on the output tensor that isn't initialized. By skipping initialization here, we're forcing users of triton-shared to generate extra code to make it correct if they want to lower linalg.matmul into loops. As a front-end, I'd think that we're responsible for generating correct code at the beginning and let other backends do certain optimization if necessary.

CaptnJackSparrow: It is the responsibility of the downstream users to expand the tt.dot op accordingly depending on the hw specific codegen requirements. Those which will require initialization will have to generate the appropriate code. Why should the top-level compiler pass assume anything about the lower level hw specific passes and generate sub-optimal code for all platforms? This pass should not care whether the code will be consumed by Maia or CPU, and as such it is the responsibility of the cpu backend to generate the necessary initialization that it needs. I am happy to make that change if required.

nhat-nguyen: >  Why should the top-level compiler pass assume anything about the lower level hw specific passes and generate sub-optimal code for all platforms? This pass should not care whether the code will be consumed by Maia or CPU, and as such it is the responsibility of the cpu backend to generate the necessary initialization that it needs.

I don't quite agree with this. If we were to write a program in a high-level language that may run on multiple platforms, it is my responsibility to make sure buffers are initialized with no assumption about the underlying behavior of the target.

I as a writer of the high-level program cannot just hope that by not initializing the buffers, somehow the compiler of the target I'm compiling to will make it right. I'm using this analogy because we have partners that use the outputs of triton-shared as their front-end and assume that the mlir code that we generate are semantically correct. And by omitting initialization here in this case, we have pushed the responsibility of correcting the code to the users of triton-shared .

I'd be inclined to think that a pass that "removes redundant zero initialization" is more useful than a pass that "looks at all uninitialized buffers and initialize them if needed because may be the user forgot to". Isn't the point of all compiler optimization is to remove unnecessary work?

PR #248, opened 2025-04-02
Do not fail pass pipeline when triton-to-unstructured fails

nhat-nguyen:
- Do not fail pass pipeline when triton-to-unstructured fails; when this analysis fails, we will fall back to raw pointer mode
- Start the analysis from int_to_ptr since they can also be sources of pointers in addition to the kernel arguments


beicy (approved):
LGTM!

PR #252, opened 2025-04-03
Handle `ptr_to_int` in triton-to-unstructured

nhat-nguyen:
This PR adds support for `ptr_to_int` in triton-to-unstructured and also fixes a bug with `int_to_ptr` 
- When we convert a pointer to an integer, materialize a new pointer using the offset that we have built up so far. This helps simplify the pointer chain leading to `ptr_to_int`.
- Only walk pointer chains originating from `int_to_ptr` if the op produces single pointers.


beicy (approved):
LGTM!

red1bluelost (approved):

PR #253, opened 2025-04-03
Add TPtrDialect

nhat-nguyen:
This PR introduces TPTrDialect. This is a temporary dialect that includes some of the ops that operate on the `!ptr.ptr` type in the proposal here: https://discourse.llvm.org/t/rfc-ptr-dialect-modularizing-ptr-ops-in-the-llvm-dialect/75142.
We will leverage this dialect to lower the remaining pointer operations that cannot be handled by `triton-to-structured` and `triton-to-unstructured` passes. Once the proposal is fully implemented, we will migrate over to using the official ops instead of this temporary dialect.

The code is borrowed from the author's draft implementation here with some simplifications:
+ https://github.com/fabianmcg/mlir-address/blob/main/include/mlir/Dialect/Address/IR/AddressOps.td
+ https://github.com/llvm/llvm-project/pull/73057/files


lib/Dialect/TPtr/IR/TPtrOps.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+void StoreOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  effects.emplace_back(MemoryEffects::Write::get(), &getAddrMutable(),
+                       SideEffects::DefaultResource::get());
+}
+
+OpFoldResult TypeOffsetOp::fold(FoldAdaptor adaptor) {
+  return adaptor.getBaseTypeAttr();
```
red1bluelost, line 46: This seems confusing too me. Since TypeOffsetOp's result is an int/index, wouldn't the fold function need to return a value or attribute of type int/index? I'm not sure how a TypeAttr would be valid to return here.

red1bluelost (approved):
LGTM. My one question I notice is also in the code which this is based on.

lib/Dialect/TPtr/IR/TPtrOps.cpp
nhat-nguyen: yeah i don't have a good answer for this. i don't see lots of usage of TypeAttr to draw my understanding from either

beicy (approved):

PR #254, opened 2025-04-07
Add option to lower triton tensor ptr ops to linalg

nhat-nguyen:
Add an option in `triton-arith-to-linalg` to convert triton ops on tensor of pointers (tt.load, tt.store, tt.bitcast, tt.int_to_ptr, tt.ptr_to_int) to linalg. This option is disabled by default and can be enabled by setting the `convert-tensor-ptr-ops` flag to true in the `triton-arith-to-linalg` pass.

This will help with lowering these triton ops to the equivalent ops in the PtrDialect later.

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -1963,10 +1963,10 @@ class AddPtrConverter : public OpConversionPattern<triton::AddPtrOp> {
         op, op->getResultTypes(), op->getOperands(), outputs, indexingMaps,
         iteratorTypes,
         [&](OpBuilder &builder, Location loc, ValueRange regionArgs) {
-          auto resultTypes = llvm::to_vector<6>(
-              llvm::map_range(op->getResultTypes(), [](Type type) {
-                return cast<TensorType>(type).getElementType();
-              }));
+          auto resultTypes = llvm::map_to_vector(
+            op->getResultTypes(), [](Type type) {
+              return cast<TensorType>(type).getElementType();
+            });
           auto *scalarOp =
               builder.create(loc, op->getName().getIdentifier(),
                              regionArgs.take_front(op->getNumOperands()),
@@ -1979,6 +1979,11 @@ class AddPtrConverter : public OpConversionPattern<triton::AddPtrOp> {
 
 // Convert triton op X operating on tensors of pointers to a linalg.generic
 // wrapping op X to operate on single pointer.
+// This pattern rewriter is almost identical to AddPtrConverter above, except
+// that the out param for the linalg op is an empty op instead of reusing one
+// of the existing operands. This is because depending on the templatized op,
+// the type of the operands might be different, so we cannot pick a default
+// operand to reuse for all cases.
 template <typename OpType>
 class TensorOpConverter : public OpConversionPattern<OpType> {
   using OpConversionPattern<OpType>::OpConversionPattern;
@@ -1986,11 +1991,10 @@ class TensorOpConverter : public OpConversionPattern<OpType> {
   LogicalResult
   matchAndRewrite(OpType op, typename OpType::Adaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto resType = op.getResult().getType();
-    if (!isa<ShapedType>(resType)) {
+    auto resultTensorType = dyn_cast<RankedTensorType>(op.getResult().getType());
+    if (!resultTensorType) {
       return failure();
     }
-    auto resultTensorType = cast<RankedTensorType>(resType);
     auto rank = resultTensorType.getRank();
     SmallVector<AffineMap> indexingMaps(
         /*numResult + numOperands*/ op->getNumResults() + op->getNumOperands(),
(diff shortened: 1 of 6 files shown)


lib/Conversion/TritonArithToLinalg/TritonArithToLinalgPass.cpp
```diff
@@ -158,11 +158,21 @@ class TritonArithToLinalgPass
     }
 
     if (addptrToLinalg) {
-      target.addDynamicallyLegalOp<triton::AddPtrOp>([](triton::AddPtrOp op) {
-        return !isa<ShapedType>(op.getResult().getType());
+      target.addDynamicallyLegalOp<triton::AddPtrOp>([](auto op) {
+        return !isa<ShapedType>(op->getOperands()[0].getType());
```
red1bluelost, line 162: I'm assuming you are adding `PtrAddOp` here when it eventually is merged. We can keep `.getResult()` here since PtrAddOp also names its results like this. There will be separate instantiations of this lambda by callee.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -1977,6 +1977,85 @@ class AddPtrConverter : public OpConversionPattern<triton::AddPtrOp> {
...
+  using OpConversionPattern<OpType>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(OpType op, typename OpType::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto resType = op.getResult().getType();
+    if (!isa<ShapedType>(resType)) {
+      return failure();
+    }
+    auto resultTensorType = cast<RankedTensorType>(resType);
```
red1bluelost, line 1993: We check `ShapedType` but then cast to `RankedTensorType`. We should pick to stick with one probably.

Also this could be a dyn_cast and put the pointer in the if-condition.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -1977,6 +1977,85 @@ class AddPtrConverter : public OpConversionPattern<triton::AddPtrOp> {
...
+        rank, utils::IteratorType::parallel);
+    SmallVector<Value> outputs = {rewriter.create<tensor::EmptyOp>(
+        op->getLoc(), resultTensorType.getShape(),
+        resultTensorType.getElementType())};
+    rewriter.replaceOpWithNewOp<linalg::GenericOp>(
+        op, op->getResultTypes(), op->getOperands(), outputs, indexingMaps,
+        iteratorTypes,
+        [&](OpBuilder &builder, Location loc, ValueRange regionArgs) {
+          auto resultTypes = llvm::to_vector(
+              llvm::map_range(op->getResultTypes(), [](Type type) {
```
red1bluelost, line 2008: You could probably use `llvm::map_to_vector` here.

test/Conversion/TritonArithToLinalg/triton_tensor_ptr_ops.mlir
```diff
@@ -0,0 +1,51 @@
...
+    %12 = tt.bitcast %8 : tensor<16x!tt.ptr<i32>> -> tensor<16x!tt.ptr<i64>>
+    %13 = tt.addptr %12, %cst : tensor<16x!tt.ptr<i64>>, tensor<16xi32>
+    %14 = tt.bitcast %13 : tensor<16x!tt.ptr<i64>> -> tensor<16x!tt.ptr<i32>>
+    %15 = arith.trunci %11 : tensor<16xi64> to tensor<16xi32>
+    tt.store %14, %15 : tensor<16x!tt.ptr<i32>>
+    tt.return
+  }
+}
+
+// CHECK: tt.addptr %in, %in_0 : !tt.ptr<i32>, i32
```
red1bluelost, line 42: I think it might be worth while to include some checks that we have generic ops.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -1977,6 +1977,85 @@ class AddPtrConverter : public OpConversionPattern<triton::AddPtrOp> {
...
+  matchAndRewrite(triton::StoreOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto storeTensorType = dyn_cast<RankedTensorType>(op.getValue().getType());
+    if (!storeTensorType) {
+      return failure();
+    }
+    auto rank = storeTensorType.getRank();
+    SmallVector<AffineMap> indexingMaps(
+        /*numResult + numOperands*/ op->getNumResults() + op.getNumOperands(),
+        rewriter.getMultiDimIdentityMap(rank));
```
red1bluelost, line 2037: Can we always assume that the inputs are contiguous? Would it be better to get the affine map associated with the layout of the operand/result.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -1977,6 +1977,85 @@ class AddPtrConverter : public OpConversionPattern<triton::AddPtrOp> {
...
+    SmallVector<utils::IteratorType> iteratorTypes(
+        rank, utils::IteratorType::parallel);
+    SmallVector<Value> outputs;
+    rewriter.replaceOpWithNewOp<linalg::GenericOp>(
+        op, op->getResultTypes(), op->getOperands(), outputs, indexingMaps,
+        iteratorTypes,
+        [&](OpBuilder &builder, Location loc, ValueRange regionArgs) {
+          auto resultTypes = llvm::to_vector<6>(
+              llvm::map_range(op->getResultTypes(), [](Type type) {
```
red1bluelost, line 2046: Same comment about `map_to_vector`

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: yeah the intermediate tensors in a triton program are always contiguous and have identity affine map. it's only during structured loads that these might not be contiguous (handled through the memref.reinterpret_cast + memref.copy case). when expanding tt.load above, we also use identity maps because the tensor of pointers are also contiguous

nhat-nguyen:
@red1bluelost thanks for the review! i addressed all the comments, let me know what you think

red1bluelost (approved):
LGTM

PR #256, opened 2025-04-08
Implement lowering of triton ops on pointer to the PtrDialect

nhat-nguyen:


beicy (approved):
LGTM!

lib/Conversion/TritonToLinalgExperimental/TritonToLinalgExperimentalPass.cpp
```diff
@@ -55,11 +57,12 @@ class TritonToLinalgExperimentalPass
     pm.addPass(createCanonicalizerPass());
 
     pm.addPass(createTritonToUnstructuredPass());
-    pm.addPass(createTritonArithToLinalgPass());
+    pm.addPass(createTritonArithToLinalgPass(true /*tensorPtrToLinalg*/));
```
red1bluelost, line 60: The LLVM style for specifying parameter names is `/*tensorPtrToLinalg=*/true`. Not sure if Triton follows that or something different. If they do follow LLVM it would be good to change.

lib/Conversion/TritonToLinalgExperimental/TritonToPtrPass.cpp
```diff
@@ -0,0 +1,487 @@
...
+      return !triton::isPtrTypeLike(ptrType);
+    });
+
+    target.addDynamicallyLegalOp<linalg::FillOp, linalg::GenericOp,
+                                 linalg::YieldOp, tensor::EmptyOp,
+                                 tensor::ExpandShapeOp, arith::SelectOp>(
+        [](auto op) {
+          return llvm::all_of(
+              llvm::concat<Value>(op->getOperands(), op->getResults()),
+              [&](Value v) { return !triton::isPtrTypeLike(v.getType()); });
```
red1bluelost, line 462: Nice use of the llvm STLExtras :)

lib/Conversion/TritonToLinalgExperimental/TritonToPtrPass.cpp
```diff
@@ -0,0 +1,487 @@
...
+  EmptyTensorConverter(const TypeConverter &typeConverter, MLIRContext *context)
+      : OpConversionPattern<tensor::EmptyOp>(typeConverter, context) {}
+
+  LogicalResult
+  matchAndRewrite(tensor::EmptyOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto newEmptyOp = rewriter.create<tensor::EmptyOp>(
+        op->getLoc(), op.getType().getShape(),
+        ptr::PtrType::get(rewriter.getContext()));
+    rewriter.replaceOp(op, newEmptyOp);
```
red1bluelost, line 83: Could use `replaceOpWithNewOp<tensor::EmptyOp>....` here instead.

lib/Conversion/TritonToLinalgExperimental/TritonToPtrPass.cpp
```diff
@@ -0,0 +1,487 @@
...
+  ExpandShapeConverter(const TypeConverter &typeConverter, MLIRContext *context)
+      : OpConversionPattern<tensor::ExpandShapeOp>(typeConverter, context) {}
+
+  LogicalResult
+  matchAndRewrite(tensor::ExpandShapeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto newExpandShapeOp = rewriter.create<tensor::ExpandShapeOp>(
+        op->getLoc(), getTypeConverter()->convertType(op.getType()),
+        adaptor.getSrc(), op.getReassociationExprs());
+    rewriter.replaceOp(op, newExpandShapeOp);
```
red1bluelost, line 104: Same about `replaceOpWithNewOp`

lib/Conversion/TritonToLinalgExperimental/TritonToPtrPass.cpp
```diff
@@ -0,0 +1,487 @@
...
+      : OpConversionPattern<arith::SelectOp>(typeConverter, context) {}
+
+  LogicalResult
+  matchAndRewrite(arith::SelectOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto newSelectOp = rewriter.create<arith::SelectOp>(
+        op->getLoc(), getTypeConverter()->convertType(op.getType()),
+        adaptor.getCondition(), adaptor.getTrueValue(),
+        adaptor.getFalseValue());
+    rewriter.replaceOp(op, newSelectOp);
```
red1bluelost, line 123: same about `replaceOpWithNewOp`

red1bluelost (approved):

test/Conversion/TritonToPtr/regular_load_store.mlir
```diff
@@ -0,0 +1,76 @@
...
+    %8 = tt.splat %a : !tt.ptr<i32> -> tensor<1024x!tt.ptr<i32>>
+    %9 = tt.addptr %8, %0 : tensor<1024x!tt.ptr<i32>>, tensor<1024xi32>
+
+    // b pointer
+    %18 = tt.splat %b : !tt.ptr<f32> -> tensor<1024x!tt.ptr<f32>>
+    %19 = tt.addptr %18, %0 : tensor<1024x!tt.ptr<f32>>, tensor<1024xi32>
+
+    %am = tt.load %9 : tensor<1024x!tt.ptr<i32>>
+
+    // cast result before doing float add
```
bmyerz0, line 18: nit: comment does not match the operation

test/Conversion/TritonToPtr/regular_load_store.mlir
```diff
@@ -0,0 +1,76 @@
+// RUN: triton-shared-opt --triton-arith-to-linalg="tensor-ptr-to-linalg"  --triton-to-ptr --canonicalize --cse %s | FileCheck %s
+
+module {
+  tt.func @kernel(%a : !tt.ptr<i32>, %b : !tt.ptr<f32>) -> () {
```
bmyerz0, line 4: to make sure I understand: we would not normally convert this IR to ptr, right? This test is for completeness in case someone is not using triton to structured memrefs?

Or am I missing something about the IR that makes it require the fallback?

test/Conversion/TritonToPtr/triton_tensor_ptr_ops.mlir
```diff
@@ -0,0 +1,132 @@
...
+// CHECK:           } -> tensor<16x!ptr.ptr>
+// CHECK:           [[VAR_21_:%.+]] = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel"]} ins([[VAR_19_]] : tensor<16xi64>) outs([[VAR_4_]] : tensor<16xi32>) {
+// CHECK:           ^bb0([[IN_25_:%.+]]: i64, [[IN_26_:%.+]]: i32):
+// CHECK:             [[VAR_22_11_:%.+]] = arith.trunci [[IN_25_]] : i64 to i32
+// CHECK:             linalg.yield [[VAR_22_11_]] : i32
+// CHECK:           } -> tensor<16xi32>
+// CHECK:           linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel"]} ins([[VAR_20_]], [[VAR_21_]] : tensor<16x!ptr.ptr>, tensor<16xi32>) {
+// CHECK:           ^bb0([[IN_27_:%.+]]: !ptr.ptr, [[IN_28_:%.+]]: i32):
+// CHECK:             [[VAR_22_12_:%.+]] = tptr.to_memref [[IN_27_]] : !ptr.ptr to memref<1xi32>
+// CHECK:             memref.store [[IN_28_]], [[VAR_22_12_]]{{.}}[[CST_0_]]{{.}} : memref<1xi32>
```
bmyerz0, line 128: Does MLIR lowering of linalg.generic handle a body with side effects okay?

test/Conversion/TritonToPtr/triton_tensor_ptr_ops.mlir
```diff
@@ -0,0 +1,132 @@
+// Test all triton ops on tensor of pointers are converted to linalg.generic
+// and all triton ops on pointers are converted to ops in the pointer dialect.
+// Original triton program:
+//    @triton.jit
+//    def tensor_ptr(in_ptr0, out_ptr0):
```
bmyerz0, line 5: it looks like we are testing these cases:
- load from i32 ptr, bitcast data to i64 to i32 ptr, load that
- i32 ptr bitcasted to i64, bitcasted to i32 ptr, store that

Will we need to test other cases? Is it possible to have a kernel argument that is already element type pointer? Is it possible to do a tt.load that returns tensor of pointer? Or are these always passed around as integers?

test/Conversion/TritonToPtr/masked_load_store.mlir
```diff
@@ -0,0 +1,49 @@
...
+    %cst = arith.constant dense<2> : tensor<16xi32>
+    %cst_0 = arith.constant dense<16> : tensor<16xi32>
+    %0 = tt.make_range {end = 16 : i32, start = 0 : i32} : tensor<16xi32>
+    %1 = tt.splat %arg0 : !tt.ptr<i32> -> tensor<16x!tt.ptr<i32>>
+    %2 = tt.addptr %1, %0 : tensor<16x!tt.ptr<i32>>, tensor<16xi32>
+    %3 = tt.load %2 : tensor<16x!tt.ptr<i32>>
+    %4 = arith.extsi %3 : tensor<16xi32> to tensor<16xi64>
+    %5 = tt.int_to_ptr %4 : tensor<16xi64> -> tensor<16x!tt.ptr<i32>>
+    %6 = arith.cmpi slt, %0, %cst_0 : tensor<16xi32>
+    %7 = tt.load %5, %6 : tensor<16x!tt.ptr<i32>>
```
bmyerz0, line 15: this is a standard boundary check mask (and maybe that's the only scenario that matters?)
but it seems like we can now deal with arbitrary masks when using ptr right? So perhaps we should test a more complex mask expression too?

test/Conversion/TritonToPtr/regular_load_store.mlir
nhat-nguyen: you're right. we would not run this IR through the fallback, and this is just for completeness. the idea is the fallback should be able to handle all triton IRs and be oblivious to what transformations have been done before.

test/Conversion/TritonToPtr/triton_tensor_ptr_ops.mlir
nhat-nguyen: thanks for asking! linalg.generic with a side-effecting body used to not be handled correctly. --canonicalize used to just delete the op and potentially the whole program. but in latest llvm upgrade, this has now been fixed. having this in linalg will also allow us to potentially fuse the ops with other linalg ops as well :)

bmyerz0: great!

test/Conversion/TritonToPtr/masked_load_store.mlir
nhat-nguyen: Yeah we can deal with arbitrary mask now. I have added another test with dynamic mask.

beicy (approved):

nhat-nguyen:
> @nhat-nguyen Hello, I have a question; I would like to ask where things like tptr.type_offset are converted off? Currently in my IR I take tptr.type_offset with me when I leave triton shared, it is not converted to the op supported in linalg, arith, memref. image

thanks for the question. the `tptr` ops are currently modeled after this proposal here: https://discourse.llvm.org/t/rfc-ptr-dialect-modularizing-ptr-ops-in-the-llvm-dialect/75142. this means that there are no lowering pipeline for these `tptr` ops. however, the plan is once the proposal is complete, we will be able to switch over to using the official `ptr` dialect and these ops can then leverage the in-tree lowering pipelines.

PR #259, opened 2025-04-10
Implement minStateScalar function to handle scalar operands in mask analysis

zhang-cx:
**Description:**
Currently, the mask analysis fails when either the left or right operand is a scalar. This PR introduces the minStateScalar function to address this issue, enabling the analysis to correctly process scalar operands.
With this patch, the analysis can now handle cases where one of the operands is a scalar or both of the operands are scalar, preventing the failure that previously occurred. As a result, tt.load can be successfully converted to tts.load.

**Changes:**
Added minStateScalar function to handle scalar operands in mask analysis
Updated mask analysis to utilize the new minStateScalar function
LIT test to test the new function

### DIFF
--- a/include/triton-shared/Analysis/MaskAnalysis.h
+++ b/include/triton-shared/Analysis/MaskAnalysis.h
@@ -90,6 +90,9 @@ private:
   LogicalResult addStates(const MaskState &lhsState, const MaskState &rhsState,
                           Location loc, OpBuilder &builder);
 
+  LogicalResult minStateScalar(const MaskState &lhsState, const MaskState &rhsState,
+                          Location loc, OpBuilder &builder);
+
   LogicalResult minStates(const MaskState &lhsState, const MaskState &rhsState,
                           Location loc, OpBuilder &builder);
   // -------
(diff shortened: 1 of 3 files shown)


nhat-nguyen:
Thank you! The new lit test is failing because we use `triton-shared-opt` instead of `%triton-opt`. Same for `FileCheck` (no `%` prefix).

nhat-nguyen:
Also would you mind adding the triton program corresponding to the IR in the lit test as comments too if you have it? That will help with readability a bit. Otherwise patch looks good to me.

zhang-cx:
> Also would you mind adding the triton program corresponding to the IR in the lit test as comments too if you have it? That will help with readability a bit. Otherwise patch looks good to me.

@nhat-nguyen  Thanks for the advice! I added the original triton function and modified the command.

nhat-nguyen:
@zhang-cx Looks good! The CLA check is still failing though. I believe you have to do `@microsoft-github-policy-service agree company="Meta"`. Your previous reply wrapped the company with square brackets so maybe that didn't work.

nhat-nguyen (approved):

zhang-cx:

PR #260, opened 2025-04-16
Fix passing pid args to call op

Salv1a:
**Description:**
The start index of `num_programs`/`program_id` arguments (missing args) is incorrect. When the args of `parentFunc` and `calleeFunc` are different, it fails (e.g. the modified test case for `tt.call` op).

**Fix:**
Modify the calculation of the start index.

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -888,8 +888,9 @@ struct CallConverter : public OpConversionPattern<triton::CallOp> {
 
           if (argsNeed > args.size()) {
             int missing = argsNeed - args.size();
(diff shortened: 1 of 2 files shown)


shengw-bd:
> thank you! would you mind just committing the actual changes and revert all the formatting changes to clean up the diffs? we also follow llvm style whenever possible, which means braces `{` are not on separate lines.

sorry, I got my clang-format mixed up. I will recommit later.

Salv1a:
> thank you! would you mind just committing the actual changes and revert all the formatting changes to clean up the diffs? we also follow llvm style whenever possible, which means braces `{` are not on separate lines.

Sorry, I got my clang-format mixed up, it's OK now.

Salv1a:
@microsoft-github-policy-service agree company="Bytedance"

nhat-nguyen (approved):

PR #263, opened 2025-04-18
Add missing TPtrIR dependency to TritonToLinalg

MikeHolman:
This pass uses TPtr, and so it should have the dependency. Not having it can sometimes cause build failures.


red1bluelost (approved):
LGTM

PR #269, opened 2025-05-07
Use memref.load/store for scalar load/store operations

python3kgae:
Currently, scalar load/store operations are lowered to affine.load/store and memref.load/store in different places.

This change aims to consolidate these operations into memref.load/store to improve consistency.


nhat-nguyen (approved):
thank you

PR #272, opened 2025-05-08
Add option to control Gather/Scatter support

python3kgae:
A new option for TritonToStructured pass was added to control gather/scatter support. 

It is enabled by default.
Use --triton-to-structured="enable-make-gather-scatter=false" to disable it.

### DIFF
--- a/include/triton-shared/AnalysisStructured/PtrAnalysis.h
+++ b/include/triton-shared/AnalysisStructured/PtrAnalysis.h
@@ -144,7 +144,7 @@ class PtrAnalysis {
       llvm::function_ref<Value(scf::ForOp op, size_t)> getReplacementVal);
 
   DenseSet<Value> maybeStructuredArgs;
-  bool enableMakeGatherScatterTensorPtr;
+  const bool enableMakeGatherScatterTensorPtr;
 
 public:
   PtrAnalysis(bool enableMakeGatherScatterTensorPtr)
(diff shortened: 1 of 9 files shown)


lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -1026,8 +1026,11 @@ LogicalResult PtrAnalysis::rewriteAddptrOp(triton::AddPtrOp op) {
...
       if (state.getRank() == 1)
         return failure();
-      auto maketptrOp = state.createTTSMakeGatherScatterTensorPtrOp(builder, op.getLoc());
-      ptrMap.map(op.getResult(), maketptrOp.getResult());
+      if (enableMakeGatherScatterTensorPtr) {
+        auto maketptrOp =
+            state.createTTSMakeGatherScatterTensorPtrOp(builder, op.getLoc());
+        ptrMap.map(op.getResult(), maketptrOp.getResult());
+      }
     }
```
nhat-nguyen, line 1034: i think this whole block should be wrapped inside if enableMakeGatherScatterTensorPtr. we should just return failure() if enableMakeGatherScatterTensorPtr is false.

include/triton-shared/AnalysisStructured/PtrAnalysis.h
```diff
@@ -144,8 +144,11 @@ class PtrAnalysis {
       llvm::function_ref<Value(scf::ForOp op, size_t)> getReplacementVal);
 
   DenseSet<Value> maybeStructuredArgs;
+  bool enableMakeGatherScatterTensorPtr;
```
nhat-nguyen, line 147: should probably make this const

lib/AnalysisStructured/PtrAnalysis.cpp

include/triton-shared/AnalysisStructured/PtrAnalysis.h
python3kgae: Done.

nhat-nguyen (approved):

PR #276, opened 2025-05-14
Support index_select style gather/scatter access.

python3kgae:
Save stride for unstructured dimension along chain of muls.
For case like
```
row_indices[:, None] * stride_m + col_offsets[None, :] * stride_n
```
When row_indices is not continuous, the offset and row for dim 0 will be
```
offset: row_indices
stride: stride_m
```
instead of
```
offset: row_indices * stride_m
stride: 1
```

For case like
```
    offs_3d_x = offs_2d_x[:, None] // Y
    offs_3d_y = offs_2d_x[:, None] % Y

    offs_3d_z = offs_2d_y[None, :]

    offs = offs_3d_x * stride_x + offs_3d_y * stride_y + offs_3d_z * stride_z
```
When dim 0 is not continuous, the offset and stride for dim 0 will be
```
offset:  (offs_2d_x // Y * stride_x) + (offs_2d_x % Y * stride_x)
stride: 1
```
Because cannot find a stride for dim 0.

Support i64 index for TTS_MakeGatherScatterTensorPtrOp.

Fix the stride issue caused by the offset and strides of subview are relative to the strided memref of the input memref while the offset and stride of reinterpret_cast are relative to the base underlying memory of the memref.

### DIFF
--- a/lib/AnalysisStructured/PtrAnalysis.cpp
+++ b/lib/AnalysisStructured/PtrAnalysis.cpp
@@ -398,7 +398,7 @@ LogicalResult PtrState::mulState(const PtrState &lhsState,
       OpFoldResult newOffset =
           mulOFRs(lhs->offsets[i], rhsStride, loc, builder);
       offsets.push_back(newOffset);
-      // Mul the scalart to stride.
+      // Mul the scalar to stride.
       OpFoldResult newStride =
           mulOFRs(lhs->strides[i], rhs->scalar, loc, builder);
       strides.push_back(newStride);
--- a/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
+++ b/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
@@ -865,8 +865,7 @@ private:
                                                   gatherOffsetElt.getResult(),
                                                   gatherDim, rewriter);
     unsigned rank = ptr.getSizes().size();
-    // Set strides to 1 for subview multiplies the existing strides with the
-    // stride of the subview.
+    // The subview should not apply an additional stride to the source.
     SmallVector<OpFoldResult> oneStrides(rank, OpFoldResult(step));
     // subview from srcPtr for mask.
     // With offsets[gatherDim] set to 0 since the offset already in
@@ -993,8 +992,7 @@ private:
       mixedDims[gatherDim] = sizes[gatherDim];
       sizes = mixedDims;
     }
-    // Set strides to 1 for subview/extract_slice multiplies the existing strides with the
-    // stride of the subview.
+    // The subview should not apply an additional stride to the source.
     SmallVector<OpFoldResult> oneStrides(rank, OpFoldResult(step));
     auto slice = rewriter.create<tensor::ExtractSliceOp>(
         loc, stVal, stValOffsets, sizes, oneStrides);
--- a/python/examples/test_index_select.py
+++ b/python/examples/test_index_select.py
@@ -65,11 +65,9 @@ def index_select_row(input_tensor, indices, dim):
 
 
 def test_index_select_row(device):
-    M, N = 8, 16  # Example sizes
-    input_tensor = torch.randn(M, N, device=device)  # Random tensor
-    indices = torch.tensor(
-        [1, 3, 5, 7], dtype=torch.int32, device=device
-    )  # Example indices
+    M, N = 8, 16
+    input_tensor = torch.randn(M, N, device=device)
+    indices = torch.tensor([1, 3, 5, 7], dtype=torch.int32, device=device)
     dim = 0  # Dimension to index along
     if device == "cpu":
         triton.runtime.driver.set_active(CPUDriver())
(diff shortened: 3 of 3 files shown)


lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -865,6 +865,9 @@ struct LoadConverter : public OpConversionPattern<tts::LoadOp> {
                                                   gatherOffsetElt.getResult(),
                                                   gatherDim, rewriter);
     unsigned rank = ptr.getSizes().size();
+    // Set strides to 1 for subview multiplies the existing strides with the
+    // stride of the subview.
```
bmyerz0, line 869: nit: I think the comment could be more general, such as

"The subview should not apply an additional stride"

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -990,8 +993,11 @@ struct StoreConverter : public OpConversionPattern<tts::StoreOp> {
       mixedDims[gatherDim] = sizes[gatherDim];
       sizes = mixedDims;
     }
+    // Set strides to 1 for subview/extract_slice multiplies the existing strides with the
+    // stride of the subview.
```
bmyerz0, line 997: same comment as above

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -398,8 +398,10 @@ LogicalResult PtrState::mulState(const PtrState &lhsState,
       OpFoldResult newOffset =
           mulOFRs(lhs->offsets[i], rhsStride, loc, builder);
       offsets.push_back(newOffset);
-      // Set stride to 1 when not continuous.
-      strides.push_back(builder.getIndexAttr(1));
+      // Mul the scalart to stride.
```
bmyerz0, line 401: typo? scalart

python/examples/test_index_select.py
```diff
@@ -0,0 +1,940 @@
...
+    )
+    return output_tensor
+
+
+def test_index_select_row(device):
+    M, N = 8, 16  # Example sizes
+    input_tensor = torch.randn(M, N, device=device)  # Random tensor
+    indices = torch.tensor(
+        [1, 3, 5, 7], dtype=torch.int32, device=device
+    )  # Example indices
```
bmyerz0, line 72: nit: I don't think these comments are helpful given the variable names

python/examples/test_index_select.py
```diff
@@ -0,0 +1,940 @@
...
+):
+    row_offsets = tl.arange(0, BLOCK_I)
+    row_indices = tl.load(indices + row_offsets * stride_i)
+
+    col_offsets = tl.arange(0, BLOCK_N)
+    input_pointers_0 = (
+        input_ptr + row_indices[:, None] * stride_m + col_offsets[None, :] * stride_n
+    )
+    data = tl.load(
+        input_pointers_0, mask=col_offsets[None, :] < (BLOCK_N // 2), other=0
```
bmyerz0, line 103: awesome this is working!

python/examples/test_index_select.py
```diff
@@ -0,0 +1,940 @@
...
+    BLOCK_I: tl.constexpr,
+    BLOCK_M: tl.constexpr,
+):
+    col_offsets = tl.arange(0, BLOCK_I)
+    col_indices = tl.load(indices + col_offsets)
+
+    row_offsets = tl.arange(0, BLOCK_M)
+    input_pointers_0 = (
+        input_ptr + row_offsets[:, None] * stride_m + col_indices[None, :]
+    )  # * stride_n
```
bmyerz0, line 241: what is this for?

python/examples/test_index_select.py
```diff
@@ -0,0 +1,940 @@
...
+        o_stride_k,
+        BLOCK_I=R,
+        BLOCK_M=M,
+        BLOCK_N=N,
+        BLOCK_K=K,
+    )
+    return output_tensor
+
+
+def test_index_select_3d(device):
```
bmyerz0, line 363: does this work for all 3 dims?

python/examples/test_index_select.py
```diff
@@ -0,0 +1,940 @@
...
+    return output_tensor
+
+
+def test_scatter_row_mask(device):
+    M, N = 8, 8  # Example sizes
+    input_tensor = torch.randn(M, N, device=device)  # Random tensor
+    indices = torch.tensor(
+        [1, 3, 5, 7], dtype=torch.int64, device=device
+    )  # Example indices
+    dst = torch.zeros((8, 16), dtype=input_tensor.dtype, device=device)
```
bmyerz0, line 606: comment here useful to say initialize to 0 so we can check the mask worked

bmyerz0: also, using an alternative like 1 or -1 could also help with debugging/catching errors, as some scenarios 0 ends up as a default value of uninitialized

bmyerz0 (commented):
Really great to see this working - I left a few suggestions

python/examples/test_index_select.py
python3kgae: Fixed.

python/examples/test_index_select.py
python3kgae: Added test for dim0/2.

python/examples/test_index_select.py
python3kgae: Removed.

python/examples/test_index_select.py
python3kgae: Removed.

lib/AnalysisStructured/PtrAnalysis.cpp

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Updated.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Updated.

nhat-nguyen (approved):

nhat-nguyen:
looks good to me. this question can be done in a follow up: now that we have the stride of the non-structured dimension, do we need to update the tts.gather_ptr op to include this information? if i remember correctly, we currently only save the strides for the structured dimensions

PR #279, opened 2025-05-21
Update triton to 7568a

nhat-nguyen:
- Fixed linker issue where `triton_libs` is missing as a dependency in triton-shared-opt
- Fixed the signature for pointer dialect attribute
- Updated triton-to-structured pass to use the default conversion infra (one-to-N is deprecated and its features have been merged to the original conversion driver)
  - After this update, fixed a bug in the update iter-arg code to consider all loop init-args because a value could be used in the loop init-args multiple times. The new mlir conversion driver code exposed this bug because now the `tts.get_structured_state` is only created once for each unique value.


enjustli:
could we merge it? 😋

PR #280, opened 2025-05-29
Support mask on gather dim for MakeGatherScatterTensorPtrOp

python3kgae:
When there's no mask, the code is like this
```
    for i in range(0, len(index_list)):
        index = index_list[i]
        do gather/scatter on index
```

When mask is immediate value, just update the upper bound of the gather scatter loop.
    The code will change to

```
    for i in range(0, min(len(index_list), mask_limit)):
        index = index_list[i]
        do gather/scatter on index
```
When mask is a dynamic value, create an scf.if to guard the gather scatter.
    The code will change to

```
    for i in range(0, len(index_list)):
        index = index_list[i]
        if i < mask_limit:
            do gather/scatter on index
```

### DIFF
--- a/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
+++ b/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
@@ -852,7 +852,8 @@ private:
       // gatherMaskDim to guard the load.
       if (auto gatherMaskDimIndex = getIntAttr(gatherMaskDim)) {
         // If the gather mask dimension is a constant, we can use it directly.
-        offsetSize = gatherMaskDimIndex.value();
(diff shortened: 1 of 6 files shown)


python3kgae:
> I think instead of handling the constant vs. dynamic case separately, we can combine them both by always generating the loop bound to be min(index_list_size, mask_bound) to simplify the code. I believe mlir is able to simplify the min(const, index_list_size) case for us automatically.

Updated.

nhat-nguyen (approved):
@python3kgae looks good. another suggestion i have but can be done in a different PR is to clean up the code in gather/scatter now that there is quite a bit of duplication.

PR #283, opened 2025-06-04
Skip i1 tensor in structured state tuple conversion

Nullkooland:
Skip tensor of i1 element type in `convertToPointerTupleWithOffsetsAndStrides` in the prepass of `TritonToStructured`, this solves pass failure in `test_accuracy_all_dims[dtype0-normal-True-1-shape0]` test case of [FlagGems all op](https://github.com/FlagOpen/FlagGems/blob/master/src/flag_gems/ops/all.py).

### DIFF
--- a/lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
+++ b/lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
@@ -90,8 +90,9 @@ public:
       // We only care about tensor of index / int (in addition to pointer type)
       // because only values of int and index type can potentially be part of a
       // pointer arithmetic sequence.
-      if (!isa<triton::PointerType>(tensorType.getElementType()) &&
-          !tensorType.getElementType().isIntOrIndex()) {
+      auto elementType = tensorType.getElementType();
+      if (isa<triton::PointerType>(elementType) ||
+          (elementType.isIntOrIndex() && !elementType.isInteger(1))) {
         // There's a subtle difference between returning failure() and
         // std::nullopt. From the documentation:
         //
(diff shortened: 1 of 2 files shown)


lib/Conversion/TritonToStructured/TritonToStructuredPass.cpp
```diff
@@ -90,8 +90,9 @@ class TritonToStructuredPass
       // We only care about tensor of index / int (in addition to pointer type)
       // because only values of int and index type can potentially be part of a
       // pointer arithmetic sequence.
-      if (!isa<triton::PointerType>(tensorType.getElementType()) &&
-          !tensorType.getElementType().isIntOrIndex()) {
+      auto elementType = tensorType.getElementType();
+      if (isa<triton::PointerType>(elementType) ||
+          (elementType.isIntOrIndex() && !elementType.isInteger(1))) {
```
python3kgae, line 95: Do we need to reverse the condition?
If we have to change order of if and else,  need to move the comment about std::nullopt as well.

Nullkooland: Moved

test/Conversion/TritonToStructured/triton-to-structured-prepass.mlir
```diff
@@ -197,3 +197,57 @@ module {
 
 // CHECK-COUNT-6: tts.get_structured_state
```
python3kgae, line 198: Could we check there's no i1 tts.get_structured_state?
Current check only check number of tts.get_structured_state, not the type.

Nullkooland: This line is from existing test.
I added `CHECK-NOT: tts.get_structured_state` at line 253 for the new test.

python3kgae: Sorry. I looked wrong lines.

python3kgae (approved):

PR #284, opened 2025-06-04
Support gather scatter with mod

python3kgae:
When add/mul 2 PtrStates, if any of the PtrState is not structured, set shape to 0 clear the mod.

This enables gather/scatter when mod result is mul with unstructured PtrState.


test/Conversion/TritonToLinalgExperimental/scatter_with_mod.mlir
nhat-nguyen, line 1: could we add a corresponding triton program that produces this IR in the test?

test/Conversion/TritonToLinalgExperimental/scatter_with_mod.mlir
```diff
@@ -0,0 +1,60 @@
...
+    %102 = arith.remsi %101, %cst : tensor<512x1xi32>
+    %103 = arith.divsi %101, %cst : tensor<512x1xi32>
+    %104 = tt.splat %16 : i32 -> tensor<512x1xi32>
+    %105 = arith.addi %103, %104 : tensor<512x1xi32>
+
+    %115 = tt.expand_dims %99 {axis = 0 : i32} : tensor<128xi32> -> tensor<1x128xi32>
+    %117 = tt.broadcast %115 : tensor<1x128xi32> -> tensor<512x128xi32>
+
+    // %0, %105, %102, %117
+    %131 = arith.muli %0, %arg19 : i32
```
nhat-nguyen: what's this comment for?

test/Conversion/TritonToLinalgExperimental/scatter_with_mod.mlir
python3kgae: Added test and with fix to make the test work.

test/Conversion/TritonToLinalgExperimental/scatter_with_mod.mlir
python3kgae: It is helper when simplify the mlir.
Removed.

lib/Analysis/OpFoldResultUtils.cpp
```diff
@@ -115,11 +118,28 @@ OpFoldResult expandOFRIndex(OpFoldResult ofr, OpFoldResult targetForTy,
...
+      for (auto dim : targetShapedTy.getShape()) {
+        shapeValues.push_back(b.create<arith::ConstantOp>(
+            loc, b.getIndexAttr(dim)));
+      }
+      RankedTensorType targetShapeTensorTy = RankedTensorType::get(
+          targetShapedTy.getShape().size(), b.getIndexType());
+      auto shapeTensor = b.create<tensor::FromElementsOp>(
+          loc, targetShapeTensorTy, shapeValues);
+      return b.create<tensor::ReshapeOp>(loc, targetTy, v, shapeTensor).getResult();
+    }
```
nhat-nguyen: i see above that you're using tt.splat for expand const to a shaped type. i'd suggest using tt.reshape instead of using the tensor.reshape to be consistent.

also this code doesn't look immediately obvious, please add a brief comment explaining that we're reshaping to the target type. also please verify that the reshape is compatible as well

lib/Analysis/OpFoldResultUtils.cpp
```diff
@@ -351,4 +371,50 @@ OpFoldResult compareOFRs(const OpFoldResult lhs, const OpFoldResult rhs,
...
+  if (lhsIntAttr) {
+    auto lhsOp =
+        b.create<arith::ConstantOp>(loc, b.getIndexAttr(lhsIntAttr.value()));
+    lhsValue = lhsOp.getResult();
+  }
+  auto rhsValue = dyn_cast<Value>(rhs);
+  if (rhsIntAttr) {
+    auto rhsOp =
+        b.create<arith::ConstantOp>(loc, b.getIndexAttr(rhsIntAttr.value()));
+    rhsValue = rhsOp.getResult();
```
nhat-nguyen: we are creating constant value with index type only to cast it back to int64 later. please create int64 value, perform the modulo, then finally cast the result to index

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -225,13 +225,33 @@ LogicalResult PtrState::addState(const PtrState &lhsState,
...
-      if (!hasConstZero(lhsState.strides[i])) {
+      if (lhsState.hasModulo()) {
+        auto mod = expandOFRIndex(lhsState.shape[i], lhsState.offsets[i], loc,
+                                  builder);
+        // Apply modulo to the offset.
+        newLhsOffset = remOFRs(lhsState.offsets[i], mod, loc, builder);
+      }
+      // When the dimension is structured, mul the offset by the stride to match
+      // the stride 1 for non-structured dimensions.
+      // If the dimension is not structured, the offset is already
```
nhat-nguyen: this section is getting very hard to understand. could you please add a sample triton code here to make things more readable

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -225,13 +225,33 @@ LogicalResult PtrState::addState(const PtrState &lhsState,
       strides.push_back(builder.getIndexAttr(1));
       // New offset is offset * stride.
       auto newLhsOffset = lhsState.offsets[i];
-      if (!hasConstZero(lhsState.strides[i])) {
+      if (lhsState.hasModulo()) {
+        auto mod = expandOFRIndex(lhsState.shape[i], lhsState.offsets[i], loc,
+                                  builder);
+        // Apply modulo to the offset.
```
nhat-nguyen: i know this code was already here. but looking at this again, it's also very hard to understand why we have to expandOFRIndex. can you please document this whole section of the code again?

lib/Analysis/OpFoldResultUtils.cpp
python3kgae: Changed to i64.

lib/AnalysisStructured/PtrAnalysis.cpp
python3kgae: Updated.

lib/AnalysisStructured/PtrAnalysis.cpp
python3kgae: Updated.

lib/Analysis/OpFoldResultUtils.cpp
python3kgae: changed to tt.reshape and added comment.

PR #286, opened 2025-06-05
Fix split pointer rewrite incorrectly handling static shape

nhat-nguyen:
Previously we assume the `shape` in `tts.make_tptr` coming out of `triton-to-structured` to contain either all 0s, or one 0 and one dynamic value. After the triton update, the value can be static as well. Fix the handling for such case.

### DIFF
--- a/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
+++ b/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
@@ -412,7 +412,7 @@ private:
     Value strideRow = ofrToIndexValue(op.getMixedStrides()[0], loc, rewriter);
     Value strideCol = ofrToIndexValue(op.getMixedStrides()[1], loc, rewriter);
 
-    Value modRow = op.getShape()[0];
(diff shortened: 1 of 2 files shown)


lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -441,23 +441,31 @@ struct MakeTensorPtrConverter
...
       casts = {cast1.getResult(), cast2.getResult()};
       wrapType = WRAP_STACKED;
-    } else {
+    } else if (isSplitDimension(parentShape[1])) {
       assert(parentShape[0] == 0);
       auto [cast1, cast2] = createSideBySideCastOps(op, adaptor, rewriter);
       casts = {cast1.getResult(), cast2.getResult()};
       wrapType = WRAP_SIDE_BY_SIDE;
+    } else {
+      assert(false && "Unexpected split pointer shape");
```
red1bluelost: Would be better to use llvm_unreachable here.

nhat-nguyen: updated :)

PR #293, opened 2025-06-27
Add reduction support for arith.mulf and arith.ori

kile01:
Enable reductions for `tt.reduce` ops with a single operation body consisting of a `arith.andi`, `arith.mulf`, `arith.muli`, `arith.xori`, or `arith.ori`.


bmyerz0 (approved):
any others that would be similarly easy to add right now? for example, arith::andi, arith::xori

nhat-nguyen (approved):

PR #294, opened 2025-07-02
Enable LLVM AddressSanitizer for triton-shared

timthlu:
Implemented support for debugging Triton programs with sanitizers:
- New MLIR pass added to triton-shared-opt: --add-debug-info, appending missing #di attributes for useful error logs
- New LLVM pass ran using opt: --sanitizer-attributes, appending missing sanitizer attributes to functions in LLVM IR
- Adding -fsanitize flags during compilation + linking. Doing so required changing llc and g++ to clang++
- Introducing environment variable SANITIZER_TYPE to enable the sanitizers during JIT compilation
- 3 scripts to facilitate simple build, runtime setup, and running a Triton program with sanitizers enabled (see usage)

Usage for anyone who wants to debug a Triton program they have written:
1. /triton_shared/scripts/build_triton_shared_for_sanitizers.sh: one time build script. Sets up a virtual environment, installs a custom LLVM with compiler-rt, openmp, clang, and mlir. Then builds triton_shared using this custom LLVM
2. source /triton_shared/scripts/setup_runtime_for_sanitizers.sh: after installation, prepare the runtime once to activate the virtual environment and set environment variables. Should be called with source when a new terminal is used
3. /triton_shared/scripts/run_triton.sh [asan|tsan] python triton_program.py: executes the Triton program with ASan or TSan enabled, with the required environment variables


scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+python3 -m venv ${CURR}/venv
+source ${CURR}/venv/bin/activate
+
+# clone triton-shared
+# needs to be cloned before building llvm for the hash
+export TRITON_PLUGIN_DIRS="${CURR}/triton_shared"
+git clone --recurse-submodules https://github.com/timthlu/triton-shared.git triton_shared
+
+cd triton_shared
+git checkout san_clean
```
bmyerz0: is this a branch?

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+cd $CURR
+
+# set up venv
+python3 -m venv ${CURR}/venv
+source ${CURR}/venv/bin/activate
+
+# clone triton-shared
+# needs to be cloned before building llvm for the hash
+export TRITON_PLUGIN_DIRS="${CURR}/triton_shared"
+git clone --recurse-submodules https://github.com/timthlu/triton-shared.git triton_shared
```
bmyerz0: can we separate the generic setup (like what is in root/README,md) from the openmp/asan-specific part of the setup?

I'm particularly concerned about the case where the user prefers dependencies such as clang, cmake, etc installed differently.

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+cd "${SOURCE_DIR}"
+
+LLVM_HASH=$(cat "${LLVM_HASH_FILE}")
+
+# CURRENT_BRANCH=$(git branch | awk '{if($1 == "*"){print $2}}')
+# if [ "$CURRENT_BRANCH" = "$BRANCH" ]; then
+#  echo "Already in ${BRANCH}, skip 'git checkout'"
+# else
+#  git checkout $BRANCH
+# fi
```
bmyerz0: debugging?

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+export CXXFLAGS="-Wno-unused-command-line-argument $CXXFLAGS" 
+export CFLAGS="-Wno-unused-command-line-argument $CFLAGS" 
+
+cd $BUILD_DIR
+cmake -GNinja -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_C_COMPILER=clang               \
+  -DCMAKE_CXX_COMPILER=clang++           \
+  -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR    \
+  -DLLVM_ENABLE_PROJECTS=$PROJECTS       \
+  -DLLVM_TARGETS_TO_BUILD="Native;NVPTX;AMDGPU" \
```
bmyerz0: are all these targets required?

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+
+echo "Installing to: $INSTALL_DIR"
+ninja -j$(nproc --all) -l$(nproc --all) || exit -1
+ninja -j$(nproc --all) -l$(nproc --all) install
+echo "Install succeeded\n"
+
+# prepare for triton_shared build
+export PATH="${INSTALL_DIR}/bin:${PATH}"
+which clang
+rm -rf ~/.triton
```
bmyerz0, line 37: is this necessary?

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+rm -rf "${TRITON_SHARED_DIR}/triton/build"
```
bmyerz0: this seems inconsistent with the fact that the script is checking out a new triton-shared

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+  -DCMAKE_C_COMPILER=clang               \
+  -DCMAKE_CXX_COMPILER=clang++           \
+  -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR    \
+  -DLLVM_ENABLE_PROJECTS=$PROJECTS       \
+  -DLLVM_TARGETS_TO_BUILD="Native;NVPTX;AMDGPU" \
+  $LLVM_SOURCE
+
+echo "Installing to: $INSTALL_DIR"
+ninja -j$(nproc --all) -l$(nproc --all) || exit -1
+ninja -j$(nproc --all) -l$(nproc --all) install
```
bmyerz0: this makes me wonder if we should have a script just for building llvm

scripts/setup_runtime_for_sanitizers.sh
```diff
@@ -0,0 +1,25 @@
...
+
+source ${CURR}/venv/bin/activate
+
+TRITON_SHARED_DIR="${CURR}/triton_shared"
+INSTALL_DIR="${CURR}/llvm-install"
+
+export PATH="${INSTALL_DIR}/bin:${CURR}/venv/bin:${PATH}"
+export LLVM_BINARY_DIR="${INSTALL_DIR}/bin"
+export TRITON_SHARED_OPT_PATH="${TRITON_SHARED_DIR}/triton/build/cmake.linux-x86_64-cpython-3.12/third_party/triton_shared/tools/triton-shared-opt/triton-shared-opt"
+# export LD_LIBRARY_PATH="${INSTALL_DIR}/lib/x86_64-unknown-linux-gnu" # for openmp
```
bmyerz0: cleanup

scripts/setup_runtime_for_sanitizers.sh
```diff
@@ -0,0 +1,25 @@
+# users need to call source setup.sh before running triton
```
bmyerz0: what is setup.sh

scripts/setup_runtime_for_sanitizers.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "Using path: $CURR"
+
+source ${CURR}/venv/bin/activate
+
+TRITON_SHARED_DIR="${CURR}/triton_shared"
+INSTALL_DIR="${CURR}/llvm-install"
+
+export PATH="${INSTALL_DIR}/bin:${CURR}/venv/bin:${PATH}"
+export LLVM_BINARY_DIR="${INSTALL_DIR}/bin"
+export TRITON_SHARED_OPT_PATH="${TRITON_SHARED_DIR}/triton/build/cmake.linux-x86_64-cpython-3.12/third_party/triton_shared/tools/triton-shared-opt/triton-shared-opt"
```
bmyerz0: do we need to hardcode the python version?

scripts/run_triton.sh
```diff
@@ -0,0 +1,21 @@
+# assume that setup was run before this
```
bmyerz0: I think this needs a usage example, do you pass it your python command?

scripts/setup_runtime_for_sanitizers.sh
```diff
@@ -0,0 +1,25 @@
+# users need to call source setup.sh before running triton
+# contains required environment variables
+
+# current directory if argument not provided
+input_path="${1:-.}"
+
+CURR=$(realpath "$input_path")
```
bmyerz0: I'm not sure about the pattern of relying on the current working directory. Seems error prone. Instead rely on an argument or required environment variable

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+LLVM_HASH_FILE="${TRITON_SHARED_DIR}/triton/cmake/llvm-hash.txt"
+if [ ! -e "${LLVM_HASH_FILE}" ]; then
+  print_error "${LLVM_HASH_FILE} does not exist"
+fi
+
+BUILD_DIR="${CURR}/llvm-build"
+INSTALL_DIR="${CURR}/llvm-install"
+SOURCE_DIR="${CURR}/llvm-project"
+LLVM_SOURCE="${SOURCE_DIR}/llvm"
+PROJECTS="clang;compiler-rt;mlir"
```
bmyerz0: can you highlight the part of the build configuration that is special compared to the LLVM triton builds?

Is it the compiler-rt being included?

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+
+python3 -m pip install --upgrade pip
+python3 -m pip install cmake==3.24 ninja pytest-xdist pybind11 setuptools torch==2.7 # need torch 2.7 for triton 3.3.0
+# sudo apt-get update -y
+sudo apt-get install -y ccache clang lld
+# pip3 install ninja cmake wheel pytest pybind11 setuptools
+
+cd "${CURR}"
+
+# build llvm
```
bmyerz0: this needs more explanation. My understanding is that you are pulling the same version of LLVM that Triton would but configuring the build differently.

scripts/setup_runtime_for_sanitizers.sh
bmyerz0: why do we need this script and run_triton to be different?

scripts/run_triton.sh
```diff
@@ -0,0 +1,21 @@
...
+# use LLVM_BINARY_DIR to obtain the locations of the .so files used for LD_PRELOAD
+
+sanitizer_type=$1
+env_args=""
+
+# echo "$*"
+
+if [ "${sanitizer_type}" = "asan" ]; then
+    env_args="LD_PRELOAD=\"$(dirname "${LLVM_BINARY_DIR}")/lib/clang/21/lib/x86_64-unknown-linux-gnu/libclang_rt.asan.so\" \
```
bmyerz0: clang version hardcoded?

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+cd $BUILD_DIR
+cmake -GNinja -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_C_COMPILER=clang               \
+  -DCMAKE_CXX_COMPILER=clang++           \
+  -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR    \
+  -DLLVM_ENABLE_PROJECTS=$PROJECTS       \
+  -DLLVM_TARGETS_TO_BUILD="Native;NVPTX;AMDGPU" \
+  $LLVM_SOURCE
+
+echo "Installing to: $INSTALL_DIR"
```
bmyerz0: this message should be specific that it is talking about llvm not triton or triton shared

scripts/build_triton_shared_for_sanitizers.sh
```diff
@@ -0,0 +1,110 @@
...
+LLVM_HASH=$(cat "${LLVM_HASH_FILE}")
+
+# CURRENT_BRANCH=$(git branch | awk '{if($1 == "*"){print $2}}')
+# if [ "$CURRENT_BRANCH" = "$BRANCH" ]; then
+#  echo "Already in ${BRANCH}, skip 'git checkout'"
+# else
+#  git checkout $BRANCH
+# fi
+
+git checkout ${LLVM_HASH}
```
bmyerz0: by the `if [ ! -e "${SOURCE_DIR}/.git" ]; then` it seems like you want to be able to run this script again. But this checkout would fail if the LLVM_HASH isn't found (e.g., due to need to git fetch)

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
```diff
@@ -0,0 +1,114 @@
+//===----------------------------------------------------------------------===//
```
bmyerz0, line 1: recommend having a lit test to unit test this code

https://llvm.org/docs/CommandGuide/FileCheck.html
https://llvm.org/docs/CommandGuide/lit.html
examples: https://github.com/microsoft/triton-shared/tree/main/test

backend/compiler.py
```diff
@@ -33,6 +34,16 @@ def _dump_ir_if_needed(files):
...
         shutil.copy(f, os.path.join(path, os.path.basename(f)))
 
+def _get_sanitizer_type():
+    # returns "" if not set
+    # throws error if set to something other than "asan"
+    sanitizer_type = os.getenv("SANITIZER_TYPE", "")
+
+    if sanitizer_type != "" and sanitizer_type != "asan":
+        # throw error
+        raise Exception(f"{sanitizer_type} is invalid.")
```
bmyerz0: Message should say the name of the env var

backend/compiler.py
```diff
@@ -33,6 +34,16 @@ def _dump_ir_if_needed(files):
     for f in files:
         shutil.copy(f, os.path.join(path, os.path.basename(f)))
 
+def _get_sanitizer_type():
+    # returns "" if not set
+    # throws error if set to something other than "asan"
+    sanitizer_type = os.getenv("SANITIZER_TYPE", "")
```
bmyerz0: consider naming this TRITON_SHARED_SANITIZER_TYPE or something like that to namespace it

lib/Sanitizer/SanitizerAttributes/SanitizerAttributes.cpp
```diff
@@ -0,0 +1,60 @@
+#include "llvm/IR/LegacyPassManager.h"
```
bmyerz0: what is the rationale to use the legacy pass manager?

backend/compiler.py
```diff
@@ -43,7 +54,16 @@ def _ttir_to_ttsharedir(mod):
...
         _dump_ir_if_needed([src_path])
         triton_shared_opt_path = _get_triton_shared_opt_path()
-        subprocess.check_call([triton_shared_opt_path, src_path, "--triton-to-linalg-experimental", "--mlir-print-debuginfo", "-o", dst_path])
+
+        subprocess_args = [triton_shared_opt_path, src_path, "--triton-to-linalg-experimental", "--mlir-print-debuginfo", "-o", dst_path]
+
+        if _get_sanitizer_type() != "":
+            print("Building with sanitizer support...")
+
+            # has to run before the other passes
```
bmyerz0: can you include why?

backend/compiler.py
```diff
@@ -118,8 +138,34 @@ def _llir_to_bin(llir: str, metadata):
...
+            subprocess.check_call([opt_path, "-load-pass-plugin", sanitizer_attributes_pass_path, 
+                "-passes=sanitizer-attributes", f"-sanitizer-type={sanitizer_type}", "-S", src_path, 
+                "-o", instrumented_src_path])
+            
+            src_path = instrumented_src_path
+
+        # compile to object file
+        clang_path = _get_llvm_bin_path("clang++")
+
+        subprocess_args = [clang_path, "-c", src_path, "-o", dst_path]
```
bmyerz0: this is now requiring clang always; can we avoid that when not using sanitizer?

backend/compiler.py
```diff
@@ -118,8 +138,34 @@ def _llir_to_bin(llir: str, metadata):
...
+        sanitizer_type = _get_sanitizer_type()
+
+        if sanitizer_type != "":
+            # using a sanitizer
+            # invoke pass to append sanitizer attributes
+            instrumented_src_path = os.path.join(tmpdir, "kernel-instrumented.ll")
+        
+            opt_path = _get_llvm_bin_path("opt")
+            top_level_triton_path = os.path.dirname(triton.__file__)
+            sanitizer_attributes_pass_path = str(next(Path(top_level_triton_path).rglob("libSanitizerAttributes.so"), None))
```
bmyerz0, line 151: what if this is None?

backend/driver.py
```diff
@@ -7,11 +7,45 @@
...
+    sanitizer_type = os.getenv("SANITIZER_TYPE", "")
+
+    if sanitizer_type != "" and sanitizer_type != "asan":
+        # throw error
+        raise Exception(f"{sanitizer_type} is invalid.")
+    
+    return sanitizer_type
+
+@lru_cache(maxsize=1)
+def _openmp_available():
```
bmyerz0: remove the threadsan specific things from this PR

backend/driver.py
```diff
@@ -271,12 +310,23 @@ def launch(
...
                   # Compile it together.
-                  subprocess.check_call([
-                    "g++", "-std=c++17", launcher_src_path, obj_path,
-                    f"-I{py_include_dir}", f"-I{include_dir}", f"-L{py_lib_dir}",
-                    "-shared", f"-l{py_lib}", "-fPIC", "-o", so_path
-                  ])
+                  clang_path = _get_llvm_bin_path("clang++")
+
+                  subprocess_args = [
+                      clang_path, "-std=c++17", launcher_src_path, obj_path,
```
bmyerz0: changing compiler path should be conditional

bmyerz0 (commented):
Awesome to see this. I took a first pass and have suggestions

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Yes, this was here for testing purposes. Once this branch is merged, the main branch should work. Removed the checkout statement and changed the repo to microsoft/triton-shared

scripts/build_triton_shared_for_sanitizers.sh
timthlu: The build script is now split into 3 scripts:
1. build_dependencies_for_sanitizers: for setting up the python virtual environment, installing clang, cmake, etc, and cloning triton_shared
2. build_llvm_for_sanitizers: building the custom LLVM (which includes ASan)
3. build triton_shared_for_sanitizers: using the custom LLVM to build triton_shared

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Removed

scripts/build_triton_shared_for_sanitizers.sh
timthlu: These are the targets supported by the Triton language. Triton's build script for LLVM also uses these exact targets, see line 5 of https://github.com/triton-lang/triton/blob/main/scripts/build-llvm-project.sh

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Yes, based on our testing, this directory needs to be removed in order to avoid LLVM linking issues in triton-shared. This is likely because the wrong LLVM build is being used in building triton-shared (due to caching), even when we give triton-shared the specific LLVM build to use with LLVM_INCLUDE_DIRS, LLVM_LIBRARY_DIR, etc.

backend/compiler.py
timthlu: The reason here is that in the AddLLVMDebugInfo pass, we have the line `moduleOp.walk([&](triton::FuncOp funcOp) {`, which only matches with functions in the triton dialect (i.e. `tt.func`). The subsequent passes translate this to other dialects, causing this pass to have no matches (and therefore no #di attributes inserted). This comment has been updated with this information.

If there is a problem with this, we can investigate changing this pass to match with other dialects.

backend/compiler.py
timthlu: Wrapped in if statement

backend/compiler.py
timthlu: This should not be `None` as long as this PR is merged, as this file is being generated as part of the changes made by this PR. If it is, this indicates a problem with the triton-shared build. Do we still need something to handle this?

backend/driver.py
timthlu: Removed

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Removed

scripts/build_triton_shared_for_sanitizers.sh
timthlu: The build script is now split into 3 scripts:
1. build_dependencies_for_sanitizers: for setting up the python virtual environment, installing clang, cmake, etc, and cloning triton_shared
2. build_llvm_for_sanitizers: building the custom LLVM (which includes ASan)
3. build triton_shared_for_sanitizers: using the custom LLVM to build triton_shared

scripts/setup_runtime_for_sanitizers.sh
timthlu: Removed

scripts/setup_runtime_for_sanitizers.sh
timthlu: It refers to the current file. Changed to be more clear

scripts/setup_runtime_for_sanitizers.sh
timthlu: Changed to recursive find

backend/driver.py
timthlu: Wrapped in if statement

lib/Sanitizer/SanitizerAttributes/SanitizerAttributes.cpp
timthlu: This pass already uses the new pass manager. This statement has been removed

backend/compiler.py
timthlu: Renamed

backend/compiler.py
timthlu: Changed

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
timthlu: Added lit test

scripts/build_triton_shared_for_sanitizers.sh
timthlu: These scripts are not meant to be ran multiple times. Removed

scripts/run_triton.sh
timthlu: Changed to recursive find

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Changed

scripts/setup_runtime_for_sanitizers.sh
timthlu: This script is meant to be run once every time the user starts up a new terminal, to set up the python virtual environment and environment variables to find LLVM and triton-shared. This script takes in multiple paths: 1) path to the venv, 2) path to LLVM install dir, and 3) path to triton shared.

The run_triton script is meant to be run every time a triton program is executed, and contains logic to enable/disable the sanitizers.

It is technically possible to combine the setup script and run script into a single script run every time a triton program is executed, but this may be inconvenient for users as they will need to input the 3 paths every time they want to run a triton program, which is why the setup script exists. Is combining them preferred?

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Yes, this is correct. We build the same version of LLVM that triton-shared currently supports, with some additional libraries for the sanitizers. This comment has been updated

scripts/build_triton_shared_for_sanitizers.sh
timthlu: Only the projects being built is different. Yes, compiler-rt contains the ASan and TSan binaries. Another new addition would be clang, which is used during compiling and linking for the flag `-fsanitize`. We build clang here manually to avoid compatibility issues between the system installed clang and our custom LLVM build.

scripts/setup_runtime_for_sanitizers.sh
timthlu: Changed to take in arguments, and give an error if any is not given

scripts/run_triton.sh
timthlu: Yes, users would run `run_triton.sh <sanitizer type> python triton_program.py`. <sanitizer type> can be "asan" or omitted if the user does not want to execute their program with a sanitizer enabled. An if statement containing usage details has been added to handle this if the user does not pass in any arguments.

timthlu:
@microsoft-github-policy-service agree company="Microsoft"

beicy (approved):

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "triton-shared/Transform/AddLLVMDebugInfo/AddLLVMDebugInfo.h"
+
+#include "llvm/BinaryFormat/Dwarf.h"
+#include "llvm/Support/Path.h"
+
+// #include "llvm/Support/raw_ostream.h"
+
```
nhat-nguyen: should remove this

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
```diff
@@ -0,0 +1,118 @@
...
+  void runOnOperation() override {
+    mlir::ModuleOp moduleOp = getOperation();
+    mlir::MLIRContext *context = &getContext();
+    mlir::Builder builder(context);
+    mlir::SymbolTable symbolTable(moduleOp);
+
+    // iterate through all functions
+    moduleOp.walk([&](triton::FuncOp funcOp) {
+      // llvm::outs() << "Found triton funcop: " << funcOp.getName() << "\n";
+
```
nhat-nguyen: remove this or make this a debug print

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
```diff
@@ -0,0 +1,118 @@
...
+      // get loc for function and pull line and file information
+      mlir::Location loc = funcOp.getLoc();
+      if (auto funcLoc = mlir::dyn_cast<mlir::FileLineColLoc>(loc)) {
+        fileName = llvm::sys::path::filename(funcLoc.getFilename().getValue());
+        filePath = llvm::sys::path::parent_path(funcLoc.getFilename().getValue());
+        line = funcLoc.getLine();
+        col = funcLoc.getColumn();
+      }
+
```
nhat-nguyen: what happens if the loc isn't a FileLineColLoc? should we signal that the pass has failed because otherwise fileName, filePath, line, col will all be empty

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
```diff
@@ -0,0 +1,118 @@
+//===----------------------------------------------------------------------===//
```
nhat-nguyen, line 1: general comment on this file: since we're using the mlir namespace, we should just remove all mlir:: prefix from all the class names / function calls

nhat-nguyen (commented):
did a quick first pass

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
```diff
@@ -0,0 +1,118 @@
...
+      if (!funcOp.isDeclaration()) {
+        subprogramFlags = mlir::LLVM::DISubprogramFlags::Definition;
+      }
+      if (isOptimized) {
+        subprogramFlags = subprogramFlags | mlir::LLVM::DISubprogramFlags::Optimized;
+      }
+      if (funcOp.getSymNameAttr() == "main") {
+        subprogramFlags = subprogramFlags | mlir::LLVM::DISubprogramFlags::MainSubprogram;
+      }
+
```
nhat-nguyen: might be worth extracting this out to a function to make this cleaner

scripts/build_dependencies_for_sanitizers.sh
```diff
@@ -0,0 +1,40 @@
+# sets up a python virtual environment, clones triton shared, and does some preliminary pip installs
```
nhat-nguyen: general comment for all other scripts: i think it might be worth it to run these as part of the daily runs to make sure these don't go stale or break in future triton releases

scripts/build_dependencies_for_sanitizers.sh
```diff
@@ -0,0 +1,40 @@
...
+python3 -m venv ${VENV_PATH}/venv
+source ${VENV_PATH}/venv/bin/activate
+
+# clone triton-shared
+export TRITON_PLUGIN_DIRS="${TRITON_SHARED_PATH}/triton_shared"
+git clone --recurse-submodules https://github.com/microsoft/triton-shared.git "${TRITON_SHARED_PATH}/triton_shared"
+
+python3 -m pip install --upgrade pip
+python3 -m pip install cmake==3.24 ninja pytest-xdist pybind11 setuptools torch==2.7 # need torch 2.7 for triton 3.3.0
+# sudo apt-get update -y
```
nhat-nguyen: should we remove these commented lines?

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
timthlu: Removed

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
timthlu: Removed

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
timthlu: We expect the Triton frontend to provide a FileLineColLoc. An else statement with signalPassFailed has been added in case the location type is different.

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
timthlu: Removed

lib/Transform/AddLLVMDebugInfo/AddLLVMDebugInfoPass.cpp
timthlu: Extracted out to a private function

scripts/build_dependencies_for_sanitizers.sh
timthlu: Removed

scripts/build_dependencies_for_sanitizers.sh
timthlu: Discussed this with Nhat in messages. These scripts require doing environment setup (venv setup, pip install) cloning triton-shared, cloning llvm-project, doing a custom build/install of LLVM, then building triton-shared. This is quite involved and time-consuming, and may not be worth to make a daily run for

scripts/build_dependencies_for_sanitizers.sh
```diff
@@ -0,0 +1,38 @@
+# sets up a python virtual environment, clones triton shared, and does some preliminary pip installs
+# in preparation for build_llvm_for_sanitizers and build_triton_shared_for_sanitizers
+# user needs to activate their venv after this
+
```
timthlu: Rename file to setup triton-shared with virtual environment

scripts/build_dependencies_for_sanitizers.sh
```diff
@@ -0,0 +1,38 @@
...
+python3 -m venv ${VENV_PATH}/venv
+source ${VENV_PATH}/venv/bin/activate
+
+# clone triton-shared
+export TRITON_PLUGIN_DIRS="${TRITON_SHARED_PATH}/triton_shared"
+git clone --recurse-submodules https://github.com/microsoft/triton-shared.git "${TRITON_SHARED_PATH}/triton_shared"
+
+python3 -m pip install --upgrade pip
+python3 -m pip install cmake==3.24 ninja pytest-xdist pybind11 setuptools torch==2.7 # need torch 2.7 for triton 3.3.0
+sudo apt-get install -y ccache clang lld
```
timthlu: Echo to user to start virtual environment: source ${VENV_PATH}/venv/bin/activate

scripts/build_dependencies_for_sanitizers.sh
```diff
@@ -0,0 +1,38 @@
...
+  echo "Error: Path '$TRITON_SHARED_PATH' does not exist."
+  exit 1
+fi
+
+# set up venv
+python3 -m venv ${VENV_PATH}/venv
+source ${VENV_PATH}/venv/bin/activate
+
+# clone triton-shared
+export TRITON_PLUGIN_DIRS="${TRITON_SHARED_PATH}/triton_shared"
```
timthlu: Remove from this script. Echo to user: exporting this is recommended

scripts/build_llvm_for_sanitizers.sh
```diff
@@ -0,0 +1,69 @@
...
+LLVM_HASH_FILE="${TRITON_SHARED_PATH}/triton/cmake/llvm-hash.txt"
+if [ ! -e "${LLVM_HASH_FILE}" ]; then
+  print_error "${LLVM_HASH_FILE} does not exist"
+fi
+
+LLVM_BUILD_DIR="${LLVM_PATH}/llvm-build"
+LLVM_INSTALL_DIR="${LLVM_PATH}/llvm-install"
+LLVM_SOURCE_DIR="${LLVM_PATH}/llvm-project"
+LLVM_SOURCE="${LLVM_SOURCE_DIR}/llvm"
+LLVM_PROJECTS="clang;compiler-rt;mlir"
```
timthlu, line 43: Comment that these projects are specifically needed to enable the sanitizers

scripts/build_llvm_for_sanitizers.sh
```diff
@@ -0,0 +1,69 @@
...
+export CXXFLAGS="-Wno-unused-command-line-argument $CXXFLAGS" 
+export CFLAGS="-Wno-unused-command-line-argument $CFLAGS" 
+
+cd "$LLVM_BUILD_DIR"
+cmake -GNinja -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_C_COMPILER=clang               \
+  -DCMAKE_CXX_COMPILER=clang++           \
+  -DCMAKE_INSTALL_PREFIX=$LLVM_INSTALL_DIR    \
+  -DLLVM_ENABLE_PROJECTS=$LLVM_PROJECTS       \
+  -DLLVM_TARGETS_TO_BUILD="Native;NVPTX;AMDGPU" \
```
timthlu: Comment why we are building these 3 targets

scripts/build_triton_shared_for_sanitizers.sh
timthlu: At the beginning of this script, check whether ~/.triton exists and error out with a message telling the user to remove it if it exists.

scripts/run_triton.sh
```diff
@@ -0,0 +1,29 @@
+# executes a triton program with sanitizers enabled/disabled
+# assume that source setup_runtime_for_sanitizers was run before this
+# use LLVM_BINARY_DIR to obtain the locations of the .so files used for LD_PRELOAD
+
+if [ "$#" -eq 0 ]; then
```
timthlu: Do not allow omission for the sanitizer type

scripts/run_triton.sh
```diff
@@ -0,0 +1,29 @@
+# executes a triton program with sanitizers enabled/disabled
```
timthlu: Rename script to be sanitizer specific. run_triton_with_sanitizer.sh

scripts/setup_runtime_for_sanitizers.sh
```diff
@@ -0,0 +1,34 @@
+# users need to call source on this file before running triton programs using run_triton
```
timthlu: Change run_triton

lib/Sanitizer/SanitizerAttributes/SanitizerAttributes.cpp
```diff
@@ -0,0 +1,59 @@
+#include "llvm/Passes/PassBuilder.h"
```
timthlu, line 1: Make a lit for this LLVM pass

backend/compiler.py
timthlu: Add some error handling to check this invariant, throw an error if it is violated

scripts/setup_runtime_for_sanitizers.sh
```diff
@@ -0,0 +1,34 @@
...
+  echo "Error: Path '$TRITON_SHARED_PATH' does not exist."
+  exit 1
+fi
+
+source ${VENV_PATH}/bin/activate
+
+export PATH="${LLVM_INSTALL_PATH}/bin:${VENV_PATH}/bin:${PATH}"
+export LLVM_BINARY_DIR="${LLVM_INSTALL_PATH}/bin"
+
+export TRITON_SHARED_OPT_PATH="$(find "$TRITON_SHARED_PATH" -type f -name "triton-shared-opt" | head -n 1)"
```
timthlu: Break this up into just the find statement. Then validate that there was exactly 1 instance found before proceeding. Throw an error if there wasn't.

Move this check as early as possible so that we don't have to undo the other statements in this file

scripts/run_triton.sh
```diff
@@ -0,0 +1,29 @@
...
+    exit 1
+fi
+
+sanitizer_type=$1
+env_args=""
+
+llvm_install_dir="$(dirname "${LLVM_BINARY_DIR}")"
+
+if [ "${sanitizer_type}" = "asan" ]; then
+    asan_dir="$(find "$llvm_install_dir" -type f -name "libclang_rt.asan.so" | head -n 1)"
```
timthlu: Check invariant here

nhat-nguyen (approved):
Thanks @timthlu. Congratulations for finishing this!

PR #297, opened 2025-07-11
Fixed permutation propagation during transpose

enjustli:


enjustli:
@nhat-nguyen

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -114,17 +114,25 @@ static SmallVector<utils::IteratorType> getNParallelLoopsAttrs(unsigned n) {
 }
 
 static Value getTransposedValue(Value source, const Location loc,
```
red1bluelost, line 119: Could you add a comment explaining the default behavior and the behavior when order is provided.

python/examples/conftest.py
```diff
@@ -18,6 +19,19 @@ def device(request):
...
+def with_allocator():
+    import triton
+    from triton.runtime._allocation import NullAllocator
+    from triton._internal_testing import default_alloc_fn
+
+    triton.set_allocator(default_alloc_fn)
+    try:
+        yield
+    finally:
+        triton.set_allocator(NullAllocator())
```
red1bluelost, line 32: What is the purpose of this fixture? Might be worth adding a comment explaining it if the fixture is essential for some test(s).

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
enjustli: - convert from triton::TransOp, it will be the triton::TransOp order
- in ReduceConverter, it will inject transpose when https://github.com/microsoft/triton-shared/blob/main/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp#L1314 condition valid

python/examples/conftest.py
enjustli: https://github.com/triton-lang/triton/blob/ec8cb09329cf25ac241a7dee1eea5a5d94daef8a/python/test/unit/language/test_core.py#L3637 
test_trans_4d needs this fixture

red1bluelost: Could you add a comment in the code file.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
red1bluelost: Could you add a comment in the code file.

enjustli: 👌

python/examples/conftest.py
enjustli: 👌

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
enjustli: already update

python/examples/conftest.py
enjustli: already update

red1bluelost (approved):
LGTM

PR #300, opened 2025-07-16
Replace memref.load/store with affine.load/store

lechenyu:
This PR modifies to generated MLIR code In ConversionPatterns.hpp. It replaces memref.laod/store with affine.load/store, and also updates lit tests


include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -525,7 +524,9 @@ struct StoreConverter : public OpConversionPattern<triton::StoreOp> {
       auto index =
```
red1bluelost, line 524: This variable is now unused

lechenyu: I will fix it in the next PR for those formatting issues.

PR #301, opened 2025-07-16
Enable LLVM ThreadSanitizer for triton-shared

timthlu:
This PR builds on top of https://github.com/microsoft/triton-shared/pull/294, which enabled LLVM AddressSanitizer for triton-shared. Enabling ThreadSanitizer support involves the same infrastructure, with some small differences:
- Requires `-fsanitize=thread` during compiling and linking
- Requires parallelizing the Triton grid using OpenMP in order to detect data races. The LLVM build script has also been changed to reflect this
- Requires some TSan-specific suppressions and libraries (archer), added to the run script

### DIFF
--- a/backend/driver.py
+++ b/backend/driver.py
@@ -114,6 +114,8 @@ extern "C" {{
 static void _launch(int gridX, int gridY, int gridZ, {arg_decls}) {{
   if (gridX*gridY*gridZ > 0) {{
     // Cast "function" to the real function type.
+    // apply parallelization to the triton grid when using ThreadSanitizer (TSan) 
+    // to help detect potential data races across program instances during kernel execution
     {"#pragma omp parallel for collapse(3)" if _get_sanitizer_type() == "tsan" else ""}
     for(int x = 0; x < gridX; x++) {{
       for(int y = 0; y < gridY; y++) {{
--- a/backend/driver.py
+++ b/backend/driver.py
@@ -20,10 +20,10 @@ def _get_llvm_bin_path(bin_name: str) -> str:
 
 def _get_sanitizer_type():
     # returns "" if not set
-    # throws error if set to something other than "asan"
+    # throws error if set to something other than "asan" or "tsan"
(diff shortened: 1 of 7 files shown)


backend/driver.py
```diff
@@ -114,6 +114,7 @@ def _generate_launcher(constants, signature, kernel_name):
 static void _launch(int gridX, int gridY, int gridZ, {arg_decls}) {{
   if (gridX*gridY*gridZ > 0) {{
     // Cast "function" to the real function type.
+    {"#pragma omp parallel for collapse(3)" if _get_sanitizer_type() == "tsan" else ""}
```
nhat-nguyen, line 119: can we add a comment here explaining the reasoning we need to parallelize when using with tsan

scripts/build_llvm_for_sanitizers.sh
```diff
@@ -39,8 +39,8 @@ LLVM_INSTALL_DIR="${LLVM_PATH}/llvm-install"
 LLVM_SOURCE_DIR="${LLVM_PATH}/llvm-project"
 LLVM_SOURCE="${LLVM_SOURCE_DIR}/llvm"
 
-# compiler-rt and clang are the sanitizer-specific LLVM projects
-LLVM_PROJECTS="clang;compiler-rt;mlir"
+# compiler-rt, clang, and openmp are the sanitizer-specific LLVM projects
+LLVM_PROJECTS="clang;compiler-rt;openmp;mlir"
```
nhat-nguyen, line 44: can we add comment that openmp is for when we're using tsan

backend/driver.py
timthlu: Done

scripts/build_llvm_for_sanitizers.sh
timthlu: Done

timthlu: Done

PR #302, opened 2025-07-17
add shape collapse for linalg broadcast/transpose/fill/reduce

enjustli:


enjustli:
@nhat-nguyen

enjustli:
@nhat-nguyen

nhat-nguyen:
thank you! would you mind explaining the rationale behind this transformation? from your examples, it's not very clear what the transformation would allow us to achieve.

enjustli:
CollapseShapePass: Purpose and Explanation

CollapseShapePass simplifies the shape representation of tensors or memrefs by collapsing higher-dimensional shapes into lower-dimensional forms, as long as the dimensions are contiguous and collapsible.

💡 Simple Analogy:

You can think of it like NumPy’s reshape() operation, but used for IR-level optimization and simplification.

🎯 Purpose:
	1.	Merges contiguous dimensions (e.g., 4D → 2D) to reduce shape complexity — this is especially important for NPU-like hardware that may not support very high-dimensional tensors (e.g., >6D).
	2.	Improves pattern matching for linalg and similar ops, which may expect tensors of certain ranks.
	3.	Simplifies analysis for downstream passes, such as loop fusion, tiling,  vectorization, or code generation.

📌 Example:

Original tensor:
```
%0 = tensor.empty(%d1, %d2, %d3) : tensor<?x?x?xf32>
```
After collapsing:
```
%1 = tensor.collapse_shape %0 [[0, 1], [2]] : tensor<?x?x?xf32> into tensor<?x?xf32>
```
Explanation: dimensions [0, 1] are merged into one dimension, resulting in a ?x? shape.

✅ Usage Scenarios:
	•	When targeting NPU-like hardware, input tensors with >6 dimensions may be unsupported and need to be collapsed.
	•	Collapsing dimensions can improve execution efficiency on such hardware, similar to the effect of collapsing nested loops.

🛠️ Technical Details:

This pass automatically matches and inserts tensor.collapse_shape or memref.collapse_shape ops, along with corresponding expand_shape ops.
If the IR includes broadcast or transpose, the pass may also update broadcast dims or permutations accordingly.

A dimension collapse will only occur if:
	•	The dimensions are contiguous
	•	Dynamic dimensions do not break the collapse “equivalence”
	•	The total number of elements remains unchanged

The op’s output may later be expand_shaped to restore the original shape to satisfy shape alignment at outputs.


🚫 Note:

This pass does not modify the data content, nor does it change the actual memory layout of the tensor.
For example, reshaping a tensor<2x4xf32> into tensor<8xf32> doesn’t change how data is laid out in memory.

lib/Conversion/TritonToLinalgExperimental/CollapseShape.cpp
```diff
@@ -0,0 +1,418 @@
...
+    return success();
+  }
+
+  LogicalResult matchAndRewrite(linalg::FillOp op,
+                                PatternRewriter &rewriter) const override {
+    assert(op->getNumResults() <= 1 && "code assumes single result!");
+    if (op->getNumResults() == 1) {
+      return collapseTensor(op, rewriter);
+    } else if (op->getNumResults() == 0) {
+      return collapseMemRef(op, rewriter);
```
nhat-nguyen, line 134: could you share examples of linalg.fill with memref inputs? if this pass runs before bufferization there shouldn't be any memref inputs to linalg.fill as far as my understanding goes

lib/Conversion/TritonToLinalgExperimental/CollapseShape.cpp
```diff
@@ -0,0 +1,418 @@
...
+    }
+    return failure();
+  }
+};
+
+// This pattern collapses a `linalg.transpose` operation that transposes a
+// tensor into a `tensor.collapse_shape` operation that collapses the tensor
+// to a smaller shape. This is useful for optimizing the performance of tensor
+// operations that involve transposing large tensors.
+struct CollapseTranspose : public OpRewritePattern<linalg::TransposeOp> {
```
nhat-nguyen: i think it will be worth pasting some snippets of the IR here so it's easier to follow (same for the other patterns)

lib/Conversion/TritonToLinalgExperimental/CollapseShape.cpp
enjustli: in this code:  
https://github.com/microsoft/triton-shared/blob/main/lib/Conversion/StructuredToMemref/StructuredToMemref.cpp#L791-L794 
it will generate memref.alloc && linalg.fill

lib/Conversion/TritonToLinalgExperimental/CollapseShape.cpp
enjustli: 👌

enjustli: already added.

enjustli:
This PR introduces an architecture-specific optimization pass that performs shape collapsing on high-rank tensors. On certain hardware platforms, such as NPUs, operators (e.g., TopK) may produce tensors with more than 6 dimensions, which exceeds the dimensionality limits of many hardware backends. Preprocessing these tensors at the IR level helps ensure compatibility and improve backend codegen robustness.

For example, the TopK operator can implicitly generate additional dimensions, resulting in overly high-rank tensors. By applying shape collapsing, we can reduce such tensors to a hardware-friendly form and avoid generating invalid IR.

Given that the goal of Triton-Shared is to lower Triton IR into community-standard MLIR dialects—enabling support for a broader range of hardware targets—this pass aligns well with that objective. It improves portability and backend compatibility, especially for architectures with limitations on tensor dimensionality.

enjustli:
> > High-rank tensors (e.g., >6D) are not supported by many accelerator backends. Therefore, applying preprocessing at the intermediate representation (IR) level helps improve backend compatibility and code generation efficiency.
> 
> i definitely agree this is a useful transformation. but currently the patterns are applied to even 2d tensors which may not be what everyone wants (see the lit tests). i'm ok with merging this pass, but maybe with an option to turn on (off by default). more preferably, we can merge this pass in but do not run it as part of triton-to-linalg. any users who want this transformation can just add this pass to run right after triton-to-linalg in their backend and it will still achieve the same effect without requiring everyone to use this transformation.

ok

enjustli:
> > > High-rank tensors (e.g., >6D) are not supported by many accelerator backends. Therefore, applying preprocessing at the intermediate representation (IR) level helps improve backend compatibility and code generation efficiency.
> > 
> > 
> > i definitely agree this is a useful transformation. but currently the patterns are applied to even 2d tensors which may not be what everyone wants (see the lit tests). i'm ok with merging this pass, but maybe with an option to turn on (off by default). more preferably, we can merge this pass in but do not run it as part of triton-to-linalg. any users who want this transformation can just add this pass to run right after triton-to-linalg in their backend and it will still achieve the same effect without requiring everyone to use this transformation.
> 
> ok

already update, add pass option, default is false. revert mlir tests change.

lib/Conversion/TritonToLinalgExperimental/TritonToLinalgExperimentalPass.cpp
```diff
@@ -52,8 +53,7 @@ class TritonToLinalgExperimentalPass
     auto moduleOp = getOperation();
     PassManager pm(&getContext(), moduleOp.getOperationName());
 
-    pm.addPass(createTritonToStructuredPass(
-        enableMakeGatherScatterTensorPtr));
+    pm.addPass(createTritonToStructuredPass(enableMakeGatherScatterTensorPtr));
```
nhat-nguyen: can we revert this style change?

enjustli: ok, but this following style is what conforms to LLVM’s code format.

nhat-nguyen (approved):

PR #304, opened 2025-07-21
MakeRangeOp needs to consider the start of the range

dbd64:
The conversion for MakeRangeOp needs to consider the start of the range. It currently assumes all ranges are constructed from 0 to the size.


red1bluelost (approved):

red1bluelost (changes requested):
Please add `test_arange` to the conftest.py which pulls tests from test_core.py

dbd64:
@microsoft-github-policy-service agree company="Microsoft"

red1bluelost (approved):

PR #310, opened 2025-07-31
Fix incorrect codegen for masks dependent on loop induction variable

nhat-nguyen:
Currently, for mask dependent on loop induction variable, we compute the mask offset by adding the mask offset *before* coming into the loop by the loop iter-arg. This is not correct when the offset has an initial value other than 0 because then the value of the offset will always be one iteration *after* the current iteration. This patch fixes the codegen and adds tests for these scenarios.

### DIFF
--- a/lib/Analysis/MaskAnalysis.cpp
+++ b/lib/Analysis/MaskAnalysis.cpp
@@ -453,6 +453,44 @@ LogicalResult MaskState::parseLoopIterArg(Value v, const Location loc,
     return failure();
   }
 
+  // This is a bit of a hack!!
+  //
+  // The offset (MaskState::start) of a mask can now depend on a loop's
+  // iter-arg like the following example:
+  //
+  // idx = offset + tl.arange(0, 4)
+  // for it in range(n):
+  //   mask = idx < size
+  //   x = tl.load(x_ptr + idx, mask=mask)
+  //   tl.store(y_ptr + idx, x, mask=mask)
+  //   idx += 4
+  //
+  // See
+  // test/Conversion/TritonToStructured/mask_loop_iter_arg.mlir and
+  // and
+  // python/examples/test_mask_loop_iter_arg.py
+  // for IR and full triton code.
+  //
+  // To support this case, we first make the following assumptions:
+  //  - MaskAnalysis is runs after PtrAnalysis's prepass finishes, which means
+  //    the offset for the load and store pointers have already been set up
+  //    at `argIndex + 1`
+  //  - The tensor of indices used by the load / store and the mask are the same
+  //    (see above where `idx` appears in both the mask and the pointer
+  //    arithmetic). This allows us to use the offset at `argIndex + 1` in the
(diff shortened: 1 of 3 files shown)


lib/Analysis/MaskAnalysis.cpp
```diff
@@ -452,32 +453,65 @@ LogicalResult MaskState::parseLoopIterArg(Value v, const Location loc,
...
+      // Make sure all ops generated for the mask state are inserted before
+      // the current loop
+      builder.setInsertionPoint(forOp);
+      if (failed(lhsState.parse(tritonValue, loc, builder))) {
+        return failure();
+      }
     }
 
+    auto dist = subOFRs(lhsState.end, lhsState.start, loc, builder);
+    this->start = forOp.getRegionIterArgs()[argIndex + 1];
```
red1bluelost: This could just use `.getRegionIterArg(argIndex + 1)`.

lib/Analysis/MaskAnalysis.cpp
```diff
@@ -452,32 +453,65 @@ LogicalResult MaskState::parseLoopIterArg(Value v, const Location loc,
...
+  //  - MaskAnalysis is runs after PtrAnalysis's prepass finishes, which means
+  //    the offset for the load and store pointers have already been set up
+  //    at `argIndex + 1`
+  //  - The tensor of indices used by the load / store and the mask are the same
+  //    (see above where `idx` appears in both the mask and the pointer
+  //    arithmetic). This allows us to use the offset at `argIndex + 1` in the
+  //    above assumption. In the future, to make this more robust, we need to
+  //    verify that the offsets are indeed the same. Or alternatively, make sure
+  //    to generate a separate start and end offset for each mask that is being
+  //    updated in loops.
```
red1bluelost, line 484: It might be better to track down the usages of mask to verify these assumptions. If it is too much to do fully correct, I think it would be better to at least error out of the compiler in cases we do not support.

For the first assumption, could it be possible that a separate unrelated iteration argument exists that would throw off the assumption that argIndex + 1 is where the offset is.

nhat-nguyen: I think making this more robust definitely something we want to do, but it may require more effort given how rare this case is (if we want to generalize this). The main problem is PtrAnalysis and MaskAnalysis are two separate analyses, and the recursive nature of these two analyses also make it hard to track certain information and verify them later. Since this hack already exists (I just added more documentation), I think we can wait until we have more programs like these to have more data on how to handle these and refactor the analyses accordingly.

As for the first assumption, as long as MaskAnalysis runs after the pre-pass, and that the mask indeed uses the same offset, argIndex + 1 will always be the offset. The pre-pass makes sure to transform the loops to have that form.

nhat-nguyen: I think what we might end up doing is to have a separate verifier pass that will ensure these assumptions and error out on supported triton IR before running these transformations. That will not require refactoring the existing code and still ensure robustness.

PR #312, opened 2025-08-07
update transpose injection when convert tt.reduce to linalg.reduce

enjustli:


red1bluelost:
> > > Could you explain the motivation for this PR? It would be good to have that in the PR description.
> > > This PR should include an executable test. If one is already present, please mention it in the PR description.
> > > From what I know, we would prefer the reduction axis to be `rank - 2`, not `0`. This is because that axis is easier to vectorize.
> > 
> > 
> > As I understand it,axis 0 can be more vectorized, tt.reduce only has one axis, so if we transpose the reduce axis to 0, the other dimensions can be collapse.
> 
> The other dimensions can be collapsed only if the memory is contiguous. A transpose that does not copy would make the memory no longer contiguous forcing us to do copies.
> 
> In general, I am not sure that Triton Shared should be doing this transpose at all. It seems to me that a transpose might be hardware dependent on whether it is beneficial.

I'm realizing that we generate the copy already with the transpose. With collapsing the shape that should be fine. Please update the PR description and add an executable test.

PR #314, opened 2025-08-07
Fix for compiler crash when analyzing loop scalar mask 

dbd64:
It was previously assumed that end start and end would be set when analyzing the mask state in a loop. When the mask state is set to scalar, compute the start and end for use in the rest of the analysis.

### DIFF
--- a/lib/Analysis/MaskAnalysis.cpp
+++ b/lib/Analysis/MaskAnalysis.cpp
@@ -235,7 +235,8 @@ LogicalResult MaskState::minStateScalar(const MaskState &lhsState,
     }
   } else {
     InFlightDiagnostic diag =
-        emitError(loc) << "Unexpected case where both lhs and rhs are not scalars";
+        emitError(loc)
+        << "Unexpected case where both lhs and rhs are not scalars";
     return failure();
   }
   return success();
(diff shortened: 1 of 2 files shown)


lib/Analysis/MaskAnalysis.cpp
```diff
@@ -507,6 +511,12 @@ LogicalResult MaskState::parseLoopIterArg(Value v, const Location loc,
       }
     }
 
+    if (!lhsState.start && !lhsState.end) {
+      assert(lhsState.scalar && "MaskState must have a scalar");
+      lhsState.start = builder.getIndexAttr(0);
+      lhsState.end = lhsState.scalar;
+    }
```
nhat-nguyen, line 518: i wonder if we should also populate `start` and `end` when `scalar` is first populated since the same situation may apply to other functions too

dbd64: The documentation for mask analysis describes them as being mutually exclusive, I'm not sure what impact that would have. A more generic way to promote from scalar is probably useful

red1bluelost: The mutual exclusion might just create a new error earlier in the mask analysis for other scenarios. Definitely something we should be looking into soon, but this fix should be good for now.

red1bluelost (approved):

lib/Analysis/MaskAnalysis.cpp
nhat-nguyen: yeah let's get this in :)

PR #315, opened 2025-08-07
Add device_assert support for tensors

timthlu:
Adds device_assert support for tensors in triton-shared
- Done by looping through every element of the tensor, extracting each element, and adding a `cf.assert` for each element
- To make asserts work in the triton-shared CPU backend: compiling asserts results in symbols that are incompatible with shared libraries - requires compilation with PIC enabled

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -792,7 +792,9 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
 
     auto assertMessage =
           llvm::formatv("Assertion `{0}` failed", op.getMessage());
-
+    
+    // The condition can only be I1 or I1Tensor (integer or tensor) from TritonOps.td.
+    // Tensors will always be RankedTensorType.
     if (isa<mlir::IntegerType>(condVal.getType())) {
       // handle scalar case
       if (!condVal.getType().isInteger(1)) {
@@ -805,21 +807,15 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
 
       rewriter.create<mlir::cf::AssertOp>(op.getLoc(), condVal,
                                           assertMessage.str());
-    } else {
+    } else if (auto tensorType = dyn_cast<RankedTensorType>(condVal.getType())) {
       // handle tensor case
-      // dimensions
-      auto tensorType = dyn_cast<RankedTensorType>(condVal.getType());
       int64_t rank = tensorType.getRank();
 
       // create identity mapping for access pattern
-      std::vector<AffineMap> indexingMaps;
-      indexingMaps.push_back(AffineMap::getMultiDimIdentityMap(rank, rewriter.getContext()));
+      SmallVector<AffineMap, 3> indexingMaps{AffineMap::getMultiDimIdentityMap(rank, rewriter.getContext())};
 
       // loops do not depend on each other
-      std::vector<utils::IteratorType> iteratorTypes;
-      for (int i = 0; i < rank; i++) {
-        iteratorTypes.push_back(utils::IteratorType::parallel);
-      }
+      SmallVector<utils::IteratorType, 3> iteratorTypes(rank, utils::IteratorType::parallel);
 
       rewriter.create<linalg::GenericOp>(
         op.getLoc(),
@@ -846,6 +842,9 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
           
           b.create<linalg::YieldOp>(loc);
         });
+    } else {
+      op.emitError("Unexpected type in triton::AssertOp");
+      return failure();
     }
 
     rewriter.eraseOp(op);
--- a/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
@@ -790,25 +790,91 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
                   ConversionPatternRewriter &rewriter) const override {
     Value condVal = op.getCondition();
 
-    if (isa<mlir::TensorType>(condVal.getType())) {
-      auto scalarVal = getScalarValue(op.getCondition(), op.getLoc(), rewriter);
-      condVal = scalarVal ? scalarVal : condVal;
-    }
-    assert(condVal && isa<mlir::IntegerType>(condVal.getType()) &&
-           "Only asserts on scalars are currently supported");
-
-    if (!condVal.getType().isInteger(1)) {
-      auto zero =
-          rewriter.create<mlir::arith::ConstantIntOp>(op.getLoc(), 0, 32);
-      auto newCond = rewriter.create<mlir::arith::CmpIOp>(
-          op.getLoc(), arith::CmpIPredicate::ne, condVal, zero);
-      condVal = newCond.getResult();
-    }
-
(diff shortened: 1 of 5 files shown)


include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -790,25 +790,91 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
...
+
+      rewriter.create<scf::ForOp>(
+        op.getLoc(), begin, dim0, step, ValueRange{}, 
+        [&](OpBuilder &b0, Location loc0, Value index0, ValueRange args0) {
+          b0.create<scf::ForOp>(
+            op.getLoc(), begin, dim1, step, ValueRange{},
+            [&](OpBuilder &b1, Location loc1, Value index1, ValueRange args1) {
+              b1.create<scf::ForOp>(
+                op.getLoc(), begin, dim2, step, ValueRange{},
+                [&](OpBuilder &b2, Location loc2, Value index2, ValueRange args2) {
```
nhat-nguyen: i think it would be easier if we lower the tt.assert to a linalg op with cf.assert in its body. that way we will have the lowering to affine.for and scf.for for free. the lowering code will be much shorter too.

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -790,25 +790,63 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
...
-      auto newCond = rewriter.create<mlir::arith::CmpIOp>(
-          op.getLoc(), arith::CmpIPredicate::ne, condVal, zero);
-      condVal = newCond.getResult();
-    }
+      rewriter.create<mlir::cf::AssertOp>(op.getLoc(), condVal,
+                                          assertMessage.str());
+    } else {
+      // handle tensor case
+      // dimensions
+      auto tensorType = dyn_cast<RankedTensorType>(condVal.getType());
```
timthlu: Is this cast guaranteed to succeed? What behavior do we want the code to have if it doesn't?

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
timthlu: Changed from scf.for to linalg.generic, which seemed to have made the code a lot cleaner (and potentially support > 3D tensors)

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: you can check this by examining the TritonOps.td file to see what constraints this op has. without having looked at it, my guess is that it is safer to check for tensor type first and put the integer case in the else clause. this is because there might be other types that aren't integer nor tensor

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -808,72 +808,44 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
...
-                  } else if (rank == 2) {
-                    element = b2.create<tensor::ExtractOp>(op.getLoc(), condVal, ValueRange{index0, index1});
-                  } else {
-                    element = b2.create<tensor::ExtractOp>(op.getLoc(), condVal, ValueRange{index0, index1, index2});
-                  }
+      // loops do not depend on each other
+      std::vector<utils::IteratorType> iteratorTypes;
+      for (int i = 0; i < rank; i++) {
+        iteratorTypes.push_back(utils::IteratorType::parallel);
+      }
```
nhat-nguyen: you can just use the constructor like so `vector<...> name(rank, utils::IteratorType::parallel)` to make the code more concise

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -808,72 +808,44 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
...
-        dim1 = rewriter.create<tensor::DimOp>(op.getLoc(), condVal, 1);
-        dim2 = rewriter.create<arith::ConstantIndexOp>(op.getLoc(), 1);
-      } else if (rank == 3) {
-        dim0 = rewriter.create<tensor::DimOp>(op.getLoc(), condVal, 0);
-        dim1 = rewriter.create<tensor::DimOp>(op.getLoc(), condVal, 1);
-        dim2 = rewriter.create<tensor::DimOp>(op.getLoc(), condVal, 2);
-      }
+      // create identity mapping for access pattern
+      std::vector<AffineMap> indexingMaps;
+      indexingMaps.push_back(AffineMap::getMultiDimIdentityMap(rank, rewriter.getContext()));
```
nhat-nguyen: can do this instead:

```
std::vector<AffineMap> indexingMaps {AffineMap::getMultiDimIdentityMap(rank, rewriter.getContext())};
```

another thing is in mlir code we usually use SmallVector instead of the std::vector. SmallVector has the same usual constructors as std::vector

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
nhat-nguyen: yes it will support arbitrary-ranked tensor now :)

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
timthlu: From TritonOps.td, it looks like the type can only be I1 or I1Tensor, so either integer or tensor. Tensors are assumed to be RankedTensorType. A comment has been added for this

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
timthlu: Changed

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
timthlu: Changed, also changed both std::vectors to SmallVector

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
```diff
@@ -790,25 +790,62 @@ struct AssertConverter : public OpConversionPattern<triton::AssertOp> {
...
+          Value element = args[0];
+          
+          // convert to I1 if needed
+          if (!element.getType().isInteger(1)) {
+            auto zero =
+                b.create<mlir::arith::ConstantIntOp>(loc, 0, 32);
+            auto newCond = rewriter.create<mlir::arith::CmpIOp>(
+                loc, arith::CmpIPredicate::ne, element, zero);
+            element = newCond.getResult();
+          }
```
nhat-nguyen: i think we can extract this out since it's identical to the integer case

nhat-nguyen (commented):
a minor nit only :)

include/triton-shared/Conversion/TritonArithToLinalg/ConversionPatterns.hpp
timthlu: We have decided to remove this since based on TritonOps.td, these should be I1 in the first place.

nhat-nguyen (approved):

PR #318, opened 2025-08-11
[MaskAnalysis] Fix implement of scalar conjunction in mask analysis.

red1bluelost:
The original implementation of conjunction with scalar mistakenly assumed that the scalar argument would use it's dimension as the mask point. This does not work because scalar arguments should be a binary and not a pivot point. To support scalar arguments, we must either use the mask argument of the conjunction or completely zero out the mask. This PR changes the code generation so that select statements embedded this pattern.

### DIFF
--- a/lib/Analysis/MaskAnalysis.cpp
+++ b/lib/Analysis/MaskAnalysis.cpp
@@ -220,24 +220,24 @@ LogicalResult MaskState::minStateScalar(const MaskState &lhsState,
                                         const MaskState &rhsState, Location loc,
                                         OpBuilder &builder) {
   if (lhsState.scalar && rhsState.scalar) {
-    dims.push_back(minOFRs(lhsState.dims[0], rhsState.dims[0], loc, builder));
-  } else if (lhsState.scalar) {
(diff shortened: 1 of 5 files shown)


lib/Analysis/MaskAnalysis.cpp
```diff
@@ -219,26 +219,34 @@ LogicalResult MaskState::addStates(const MaskState &lhsState,
...
         emitError(loc)
         << "Unexpected case where both lhs and rhs are not scalars";
     return failure();
   }
+
+  // If we see a scalar condition in a conjunction with a mask, this means we
+  // are either going to take the mask dimension or take nothing at all. To do
+  // that we use a select on the scalar value with the mask dimension in the
+  // true case and zero in the false case.
+  auto &scalarState = lhsState.scalar ? lhsState : rhsState;
```
nhat-nguyen: i think this makes sense to me, but for others who aren't as familiar, having a triton snippet here would help. otherwise everything else looks good to me.

nhat-nguyen (approved):

nhat-nguyen (approved):

PR #319, opened 2025-08-12
Unify tts.gather/scatter with tts.make_gather_scatter_tptr

python3kgae:
This change merges 2 groups of gather/scatter representations in tts dialect.
```
tts.gather %ptr[%offset] %mask %other
```
is replaced with
```
gather_ptr = tts.make_gather_scatter_tptr %ptr gather_scatter_dim: 0 gather_scatter_offset: %offset gather_scatter_mask:%mask
tts.load %gather_ptr static_mask_dims = array<i64: 0>  %other
```

```
tts.scatter %ptr %offset %value  %mask
```
is replaced with
```
%scatter_ptr = tts.make_gather_scatter_tptr %ptr gather_scatter_dim: 0 gather_scatter_offset: %offset gather_scatter_mask:%mask
tts.store %scatter_ptr %value static_mask_dims = array<i64: 0>
```
Scalar load/store will use tts.load/store.

Since the %offset is always 1D, to match the result type for load and store value type for store, tensor.collapse_shape/expand_shape op will be generated.

Optional operand %gather_scatter_mask is added to tts.make_gather_scatter_tptr to save bool like mask. This is because  tts.load/store only support size mask.
%gather_scatter_mask will be used to support case where non-continuous dimension has mask which fails in MaskAnalysis and masks in other dimension pass MaskAnalysis.

The convert from tts.gather/scatter to memref is moved to structured to memref pass. And unstructured to memref pass is removed.


include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
```diff
@@ -120,7 +120,7 @@ def TTS_MakeTensorPtrOp
   //let hasCanonicalizer = 1;
 }
 
-def TT_IndexTensorLike : AnyTypeOf<[I32Tensor, I64Tensor]>;
+def TT_IndexTensorLike : AnyTypeOf<[TT_IntLike, I16Tensor, I32Tensor, I64Tensor]>;
```
red1bluelost: Why is IntLike being included here? Int does not includes non-tensor types.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -833,9 +847,121 @@ struct LoadConverter : public OpConversionPattern<tts::LoadOp> {
...
+                          .getResult();
+
+    auto baseTensor =
+        rewriter
+            .create<bufferization::ToTensorOp>(
+                loc,
+                RankedTensorType::get(
+                    SmallVector<int64_t>(1, ShapedType::kDynamic),
+                    resultType.getElementType()),
+                baseMemref, true /* restrict */, false /* writable */)
```
red1bluelost: The LLVM style for adding a name to a constant parameter is `/*restrict=*/true`.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -1142,11 +1395,55 @@ struct StoreConverter : public OpConversionPattern<tts::StoreOp> {
...
+  using OpConversionPattern<triton::StoreOp>::OpConversionPattern;
+
+public:
+  ScalarStoreConverter(const TypeConverter &typeConverter, MLIRContext *context)
+      : OpConversionPattern<triton::StoreOp>(typeConverter, context) {}
+
+  LogicalResult
+  matchAndRewrite(triton::StoreOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    assert(!op.getMask());
```
red1bluelost: Are we certain that mask will never be present? I think we should add a test where a scalar store is done with a mask since I suspect this might be possible to appear.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -975,6 +1100,50 @@ struct LoadConverter : public OpConversionPattern<tts::LoadOp> {
...
+  using OpConversionPattern<triton::LoadOp>::OpConversionPattern;
+
+public:
+  ScalarLoadConverter(const TypeConverter &typeConverter, MLIRContext *context)
+      : OpConversionPattern<triton::LoadOp>(typeConverter, context) {}
+
+  LogicalResult
+  matchAndRewrite(triton::LoadOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    assert(!op.getMask());
```
red1bluelost: Same concern about whether a mask would be present.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
python3kgae: This is to support scalar case where the offset is a scalar.
Another way to support scalar load/store is to keep the tt.add_ptr in `triton-to-unstructured` pass for scalar tt.load/store. Then convert them later in `triton-ptr-to-memref` pass. That will require clone tt.add_ptr when it is shared between tt.load/store and other operations like tt.splat.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: This is for scalar load.
Mask should not be present.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: I see. I'll add test for scalar load/store with mask.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Maybe I was wrong. I'll add a test to see if it is possible.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Fixed.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Test added.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: You're right. Added the support.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
red1bluelost: Should be no spaces and have the equal sign. https://llvm.org/docs/CodingStandards.html#comment-formatting

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
red1bluelost: If that is the case, maybe we should rename this to TTS_OffsetLike or some other more accurate name.

I also suspect that the prefix should be TTS instead of TT since we are not the Triton repo.

red1bluelost: Also, TT_IntLike is likely too general. The definition includes I1, I4, I8, I16, I32, I64 and all the tensor versions of these integers. If it is not too general, we might as well replace TT_IndexTensorLike usages with TT_IntLike.

lib/Conversion/StructuredToMemref/StructuredToMemrefPass.cpp
```diff
@@ -94,6 +94,15 @@ class StructuredToMemrefPass
...
 
+    target.addDynamicallyLegalOp<triton::LoadOp>([&](triton::LoadOp op) {
+      return !isa_and_present<tts::MakeGatherScatterTensorPtrOp>(
+          op.getPtr().getDefiningOp());
+    });
+    target.addDynamicallyLegalOp<triton::StoreOp>([&](triton::StoreOp op) {
+      return !isa_and_present<tts::MakeGatherScatterTensorPtrOp>(
+          op.getPtr().getDefiningOp());
+    });
+
```
nhat-nguyen: why is a load/store op now taking the result of tts::make_gather_scatter_tptr?

nhat-nguyen: i don't like how we are mixing results of tts ops with the original triton ops now. also, this pass is intended to lower tts ops only, but now we're lowering triton ops as well. i'm assuming this is due to the removal of the 1d rank check in ptr-analysis?

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Fixed more places.

include/triton-shared/Dialect/TritonStructured/IR/TritonStructuredDialect.td
python3kgae: Renamed to TTS_OffsetLike  and limited to 16,32 and 64 bits.

test/Conversion/StructuredToMemref/gather_mask_no_other.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: triton-shared-opt --triton-to-unstructured --canonicalize --unstructured-to-memref --canonicalize %s | FileCheck %s
+// RUN: triton-shared-opt --triton-to-unstructured --canonicalize  --structured-to-memref --canonicalize %s | FileCheck %s
```
nhat-nguyen, line 1: this pass pipeline doesn't make sense anymore. previously we can run triton-to-unstructured which produces tts.gather/scatter (unstructured ops). but now looking at this, it's not obvious why we're having to run structured-to-memref. i would rather we keep the original unstructured-to-memref (or rename it to triton-load-store-to-memref) and have them lower scalar load store instead of putting them all in structured-to-memref

test/Conversion/StructuredToMemref/reducemax_32_256_bf16.mlir
```diff
@@ -41,12 +41,12 @@ module {
...
 
-// CHECK-DAG:   [[MAP_0_:#.+]] = affine_map<(d0, d1) -> (d0, d1)>
+// CHECK-DAG:   [[MAP_0_:#.+]] = affine_map<(d0) -> (d0)>
 // CHECK-LABEL:  func.func @kernel
 // CHECK-SAME:   ([[PARAM_0_:%.+]]: memref<*xbf16>, [[PARAM_1_:%.+]]: memref<*xbf16>, [[PARAM_2_:%.+]]: i32, [[PARAM_3_:%.+]]: i32, [[PARAM_4_:%.+]]: i32, [[PARAM_5_:%.+]]: i32, [[PARAM_6_:%.+]]: i32, [[PARAM_7_:%.+]]: i32) {
 // CHECK-DAG:       [[CST_0_:%.+]] = arith.constant 0 : i32
-// CHECK-DAG:       [[VAR_empty_offsets_:%.+]] = tensor.empty() : tensor<256x16xi32>
-// CHECK-DAG:       [[VAR_zero_offsets_:%.+]] = linalg.fill ins([[CST_0_]] : i32) outs([[VAR_empty_offsets_]] : tensor<256x16xi32>) -> tensor<256x16xi32>
+// CHECK-DAG:       [[VAR_empty_offsets_:%.+]] = tensor.empty() : tensor<4096xi32>
+// CHECK-DAG:       [[VAR_zero_offsets_:%.+]] = linalg.fill ins([[CST_0_]] : i32) outs([[VAR_empty_offsets_]] : tensor<4096xi32>) -> tensor<4096xi32>
```
nhat-nguyen, line 49: why are these being flattened out?

nhat-nguyen:
> Optional operand %gather_scatter_mask is added to tts.make_gather_scatter_tptr to save bool like mask. This is because tts.load/store only support size mask.

this will make things very confusing because the mask information can now be present in either load/store or the tts.make_tptr op. it will also complicate the lowering of the tts.load/store and also the construction of the tts.make_tptr op. one of the things that this codebase suffers from is trying to use a single concept for many things which make maintenance pretty hard. i know we briefly discussed about unifying the two representations, but looking at the implementation, it feels very forced and maybe these two concepts are better off having their own representation (perhaps we should rename tts.gather/scatter to something else to better distinguish it from the tts.make_gather_scatter use case).

lib/Conversion/StructuredToMemref/StructuredToMemrefPass.cpp
python3kgae: No. It is to support scalar load/store.

I can change it to keep the tt.add_ptr in triton-to-unstructured pass for scalar tt.load/store. Then convert them later in triton-ptr-to-memref pass. That will require clone tt.add_ptr when it is shared between tt.load/store and other operations like tt.splat.
Does that work for you?

test/Conversion/StructuredToMemref/reducemax_32_256_bf16.mlir
python3kgae: Because tts.make_gather_scatter_tptr requires index list size match the size of the gather_scatter dimension for the result/storeValue type.
When we have the flattened index list, have to flatten the result/storeValue.

python3kgae:
> > Optional operand %gather_scatter_mask is added to tts.make_gather_scatter_tptr to save bool like mask. This is because tts.load/store only support size mask.
> 
> this will make things very confusing because the mask information can now be present in either load/store or the tts.make_tptr op. it will also complicate the lowering of the tts.load/store and also the construction of the tts.make_tptr op. one of the things that this codebase suffers from is trying to use a single concept for many things which make maintenance pretty hard. i know we briefly discussed about unifying the two representations, but looking at the implementation, it feels very forced and maybe these two concepts are better off having their own representation (perhaps we should rename tts.gather/scatter to something else to better distinguish it from the tts.make_gather_scatter use case).
> 
> my understanding is we wanted to do this because we want to reduce complexity in the internal structured-to-maia pass of having to support additional tts.gather/scatter. but now with the added mask difference, i don't think whatever complexity we reduced by avoiding lowering tts.gather/scatter is now lost by having to force the representation of tts.gather/scatter. we're also having more complexity in the lowering path for scalar load/store.

The reason to add %gather_scatter_mask is not only to unify tts.gather/scatter. It will also be used  to support pattern where the mask on gather_scatter_dim is fail in MaskAnalysis. In that case, it is still row continuous, but with a generic mask for gather scatter dimension.
The mask on tts.load/store will be sizeMask where %gather_scatter_mask will be for case the mask is a generic mask.

Or we could allow tts.load/store support both sizeMask and generic mask. But we'll only allow generic mask on gather scatter dimension. So add it to tts.load/store is not good as well.

Let me know if you have better suggestion.

test/Conversion/StructuredToMemref/gather_mask_no_other.mlir
python3kgae: How about lower scalar load store in triton-ptr-to-memref pass?

lib/Conversion/StructuredToMemref/StructuredToMemrefPass.cpp
red1bluelost: I think that cloning tt.add_ptr sounds fine. It seems like it'll clean up in our conversions plus CSE. Alternatively, we probably should not be removing tt.add_ptr if it a separate user.

nhat-nguyen: @python3kgae can you elaborate on why we need to clone tt.addptr? shouldn't the triton-to-unstructured already convert tt.addptr to arith.addi for tensor of offsets already?

python3kgae: Now we want to keep using tt.addptr as ptr for tt.load/tt.store so no tts and tt operations are mixed.
In triton-to-unstructured pass, we want to leave scalar tt.addptr + tt.load/tt.store for later pass.

If we see scalar tt.load/store, just skip the conversion.
If we see scalar tt.addptr, we want to skip the conversion for the tt.load/store uses. But it might be used for things like tt.splat as well. 

In case a scalar tt.addptr used by both tt.load/store and other operations.
We'll need to keep the tt.addptr for tt.load/tt.store, but we'll need to convert it for other operations like tt.splat at the same time. If that happens, we'll need to clone the addptr.

Here's the code to do the job.
```
      if (auto ptrType = dyn_cast<triton::PointerType>(addptr.getType())) {
        // Leave scalar ptrs for TritonPtrToMemrefPass to handle.
        if (ptrType.getPointeeType().isSignlessIntOrIndexOrFloat()) {
          // Collect all load/store users of the addptr op.
          // And check if there's other users of the addptr.
          SmallVector<OpOperand> ldStUsers;
          bool hasOtherUsers = false;
          for (auto user : addptr->getUsers()) {
            if (isa<triton::LoadOp, triton::StoreOp>(user)) {
              ldStUsers.push_back(user);
            } else {
              hasOtherUsers = true;
            }
          }
          if (hasOtherUsers) {
            if (!ldStUsers.empty()) {
              // Clone the addptr op and move the load/store users to
              // the cloned op.
              OpBuilder b{addptr};
              auto loc = addptr->getLoc();
              auto clone = b.create<triton::AddPtrOp>(
                  loc, addptr.getType(), addptr.getPtr(),
                  addptr.getOffset());
              for (auto &ldStUser : ldStUsers) {
                ldStUser.assign(clone);
              }
            }
            // Continue convert current addptr which get load/store users removed.
          } else {
            // Skip if only have load/store users.
            return success();
          }
        }
      }
```

nhat-nguyen: i still don't quite get it. cloning the addptr defeats the purpose of the triton-to-unstructured pass because this pass is supposed to remove all the intermediate tt.addptr if possible such that the only ops we generate are the arith.addi to compute the offset. can you please share an IR example?

python3kgae: My current plan is to lower scalar tt.load/store and the addptr for these load/store in later pass.
triton-to-unstructured pass doesn't convert these scalar tt.load/store anymore.
IR like this will be the same before and after triton-to-unstructured pass.

```
module {
  tt.func public @masked_scalar_gather_scatter(%arg0: !tt.ptr<i32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<i32> {tt.divisibility = 16 : i32}, %arg2: i1, %arg3: i1, %arg4: i32) attributes {noinline = false} {
    %c99_i32 = arith.constant 99 : i32
    %p = tt.addptr %arg0, %arg4 : !tt.ptr<i32>, i32
    %0 = tt.load %p, %arg2, %c99_i32 : !tt.ptr<i32>
    tt.store %arg1, %0, %arg3 : !tt.ptr<i32>
    tt.return
  }
}
```

Then these scalar load/store will be lowered in TritonPtrToMemrefPass where we convert the triton function signatures.

nhat-nguyen: But wouldn't that mean duplicating the work of triton-to-unstructured? How are you going to lower these ops in later pass?

nhat-nguyen: what if we have this:

```
module {
  tt.func public @masked_scalar_gather_scatter(%arg0: !tt.ptr<i32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<i32> {tt.divisibility = 16 : i32}, %arg2: i1, %arg3: i1, %arg4: i32) attributes {noinline = false} {
    %c99_i32 = arith.constant 99 : i32
    %p = tt.addptr %arg0, %arg4 : !tt.ptr<i32>, i32
    %p1 = tt.addptr %p1, %arg4 : !tt.ptr<i32>, i32
    %p2 = tt.addptr %p2, %arg4 : !tt.ptr<i32>, i32
    %0 = tt.load %p2, %arg2, %c99_i32 : !tt.ptr<i32>
    tt.store %arg1, %0, %arg3 : !tt.ptr<i32>
    tt.return
  }
}
```

previously triton-to-unstructured will collapse all the intermediate tt.addptr using the offset info and lower them to just offset addition. are we going to reuse this logic?

python3kgae: I see. Then I should always create a new addptr for scalar load/store in triton-to-unstructured as long as the ptr of addptr is another addptr to support the collapse case.

python3kgae: Tried to always created a new addptr for scalar load/store.
But that will make things like scf.if with tt.ptr result to triton-ptr-to-memref pass which require bigger change :(

To avoid current PR get even bigger, I have 2 plans for scalar load/store now:

1. Add tts.scalar_load/store or just keep tts.gather/scatter to do scalar load/store only.

2. Allow tts.load/store to support scalar result/storeValue.

Let me know if you have other suggestions for scalar load/store.

nhat-nguyen:
@python3kgae could you also help me understand how the mask tensor for the gather dimension work? if we have a 2d load, the mask tensor is going to be 2d. but the gather dimension will want a 1d tensor. how do we extract this out of the original 2d mask tensor?

lib/Conversion/StructuredToMemref/StructuredToMemrefPass.cpp
python3kgae: Switched to tts.load/store for scalar load/store.

test/Conversion/StructuredToMemref/reducemax_32_256_bf16.mlir
bmyerz0: How do we deal with the fact that the tts.load result is also flattened shape?

In examples like this it just does store so doesn't matter
```
// CHECK:               %[[VAL_36:.*]] = tensor.collapse_shape %[[VAL_35]] {{.*}}0, 1{{.*}} : tensor<2x2xi32> into tensor<4xi32>
// CHECK:               %[[VAL_37:.*]] = tts.make_gather_scatter_tptr %[[VAL_0]] to sizes: [4] gather_scatter_dim: 0 gather_scatter_offset: %[[VAL_36]], strides: [1], offsets: [0] : tensor<4xi32>  <f32> to !tt.ptr<tensor<4xf32>>
// CHECK:               %[[VAL_38:.*]] = "tts.load"(%[[VAL_37]]) <{operandSegmentSizes = array<i32: 1, 0, 0>, static_mask_dims = array<i64>}> : (!tt.ptr<tensor<4xf32>>) -> tensor<4xf32>
// CHECK:               %[[VAL_39:.*]] = tensor.collapse_shape %[[VAL_31]] {{.*}}0, 1{{.*}} : tensor<2x2xi32> into tensor<4xi32>
// CHECK:               %[[VAL_40:.*]] = tts.make_gather_scatter_tptr %[[VAL_1]] to sizes: [4] gather_scatter_dim: 0 gather_scatter_offset: %[[VAL_39]], strides: [1], offsets: [0] : tensor<4xi32>  <f32> to !tt.ptr<tensor<4xf32>>
// CHECK:               "tts.store"(%[[VAL_40]], %[[VAL_34]]) <{static_mask_dims = array<i64>}> : (!tt.ptr<tensor<4xf32>>, tensor<4xf32>) -> ()
```

But are there other cases where we need the shape?

test/Conversion/TritonToUnstructured/gather_simple_loop.mlir
```diff
@@ -34,5 +34,7 @@ module {
 // CHECK-NOT: tt.load
 // CHECK-NOT: tt.store
 
-// CHECK: [[tensor:%.+]] = tts.gather %arg0
-// CHECK: tts.scatter [[tensor]] into %arg1
+// CHECK:             %[[PTR:.*]] = tts.make_gather_scatter_tptr %arg0 to sizes: [64] gather_scatter_dim: 0 gather_scatter_offset: {{.*}}, strides: [1], offsets: [0] : tensor<64xi32>  <f32> to !tt.ptr<tensor<64xf32>>
```
bmyerz0: For these loop lit tests, should the CHECK should include how the pointers get incremented?

python/examples/test_gather_scatter.py
```diff
@@ -161,3 +161,41 @@ def test_complex_gather_scatter(device):
...
+    output = torch.tensor([-1], device=device, dtype=torch.int32)
+
+    if device == 'cpu':
+        triton.runtime.driver.set_active(CPUDriver())
+
+    grid = lambda meta: (1,)
+
+    masked_scalar_gather_scatter[grid](input, output, in_mask=False, out_mask=False)
+    
+    expected_output = torch.tensor([-1], device=device, dtype=torch.int32)
```
bmyerz0: Can these checks be a bit more robust in some way? The kernel unfortunately has the same effect as not running it at all.

One simple idea is to have a sentinel scalar that always gets written to

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -827,15 +841,180 @@ struct LoadConverter : public OpConversionPattern<tts::LoadOp> {
...
+          if (!op.hasMask()) {
+            // If there is no mask, simply extract the current element from the
+            // base tensor and use it as the yield value.
+            auto loadValue = getValueAtIndex(b, loc, offset);
+            b.create<linalg::YieldOp>(loc, loadValue);
+          } else {
+            // If the mask value is truthy, the current element is loaded from
+            // the base tensor using its offset. Otherwise, if `other` is
+            // present, yield `other`. If `other` is not present, a default
+            // value of 0 is used.
```
bmyerz0, line 973: why do we need a default value

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -827,15 +841,180 @@ struct LoadConverter : public OpConversionPattern<tts::LoadOp> {
...
+                              loc,
+                              RankedTensorType::get(
+                                  SmallVector<int64_t>(1, ShapedType::kDynamic),
+                                  resultType.getElementType()),
+                              baseMemref, /*restrict=*/true, /*writable=*/false)
+                          .getResult();
+
+    // The linalg.generic op should have the following inputs:
+    // - the offset tensor.
+    // - an optional mask tensor if the gather op contains mask.
```
bmyerz0: nit: gather op -> load op?

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -224,6 +225,19 @@ static void fillWithValue(Location loc, Value alloc, Value other,
...
 }
 
+static MemRefType getMemrefTypeForScalarPtr(triton::PointerType ptrType,
+                                            MLIRContext *context) {
+  SmallVector<int64_t> strides{1};
+  auto layout = StridedLayoutAttr::get(context, ShapedType::kDynamic, strides);
+  auto elemType = ptrType.getPointeeType();
+  if (auto shapedType = dyn_cast<ShapedType>(elemType)) {
+    // If the pointee type is a shaped type, we need to get its element type.
+    elemType = shapedType.getElementType();
```
bmyerz0: do you need an assertion that the shapedType is unit size?

test/Conversion/StructuredToMemref/gather_mask_no_other.mlir
bmyerz0: using ptr to memref seems intuitive. But will this affect triton-shared users who want to support scalar loads without full tensor of ptr support?

test/Conversion/StructuredToMemref/reducemax_32_256_bf16.mlir
python3kgae: The load result should be expand back.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: My guess is to explicit modeling of undefined or uninitialized behavior.

@nhat-nguyen Do you have more details about this default value?

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Fixed.

python/examples/test_gather_scatter.py
python3kgae: Updated.

test/Conversion/TritonToUnstructured/gather_simple_loop.mlir
python3kgae: Added.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Good catch. Removed the if.

test/Conversion/StructuredToMemref/gather_mask_no_other.mlir
python3kgae: Now scalar loads will use tts.load/store as well.
As long as triton-to-unstructured is enabled, scalar load/store should be converted to memref in --structured-to-memref.

With tts.load/store support scalar types, lower all cases of tts.load/store in structured-to-memref seems reasonable.

bmyerz0 (approved):

PR #320, opened 2025-08-12
Adds flag for old transpose behavior on reduction conversion.

red1bluelost:
Preferably Triton Shared should not be doing any transposes for optimizations because we lack the wider analysis to understand if these transformations are going to be helpful or not. That will be a part of future work.

For now, we need the old transpose behavior on the Reduce Conversion where it just transposes rank - 1 reductions to rank - 2. This is added with a flag that is default off.

### DIFF
--- a/include/triton-shared/Conversion/TritonArithToLinalg/Passes.td
+++ b/include/triton-shared/Conversion/TritonArithToLinalg/Passes.td
@@ -16,6 +16,10 @@ def TritonArithToLinalg : Pass<"triton-arith-to-linalg", "mlir::ModuleOp"> {
              "Convert tt.assert to cf.assert">,
       Option<"tensorPtrToLinalg", "tensor-ptr-to-linalg", "bool", /*default*/"false",
              "Convert triton ops on tensor of pointers to linalg.generic">,
+      Option<"transposeReduceToRank0", "transpose-reduce-to-rank0", "bool", /*default*/"true",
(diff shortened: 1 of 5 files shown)


include/triton-shared/Conversion/TritonArithToLinalg/Passes.td
```diff
@@ -16,6 +16,10 @@ def TritonArithToLinalg : Pass<"triton-arith-to-linalg", "mlir::ModuleOp"> {
              "Convert tt.assert to cf.assert">,
       Option<"tensorPtrToLinalg", "tensor-ptr-to-linalg", "bool", /*default*/"false",
              "Convert triton ops on tensor of pointers to linalg.generic">,
+      Option<"transposeReduceToRank0", "transpose-reduce-to-rank0", "bool", /*default*/"true",
+             "Transpose reductions to rank 0 which collapses remaining dimensions. Otherwise "
```
nhat-nguyen, line 20: this is default true but the pr description says default off. should we have it on or off?

red1bluelost: This is `true` as in the new behavior. The old behavior is `false`. Hence old behavior is off by default.

red1bluelost: Changed the description to make it less ambiguous.

nhat-nguyen (approved):

PR #321, opened 2025-08-14
Enable 1D tensor support for enableMakeGatherScatterTensorPtr

python3kgae:
This will allow the sized mask be collect on tts.load/tts.store for later pass. 
Without the change, tts.gather/tts.scatter will be generated with a bool tensor mask.

### DIFF
--- a/test/Conversion/TritonToStructured/gather_scatter_ptr_1d_with_mask.mlir
+++ b/test/Conversion/TritonToStructured/gather_scatter_ptr_1d_with_mask.mlir
@@ -1,0 +1,40 @@
+// RUN: triton-shared-opt --triton-to-structured --remove-dead-values --canonicalize --cse %s | FileCheck %s
+
+// Make sure tts.load is generated with correct mask for 1D tensor.
+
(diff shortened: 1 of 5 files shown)


test/Conversion/TritonToStructured/gather_scatter_ptr_1d_with_mask.mlir
```diff
@@ -0,0 +1,40 @@
+// RUN: triton-shared-opt --triton-to-structured --remove-dead-values --canonicalize --cse %s | FileCheck %s
```
bmyerz0, line 1: is there a similar test with other control flow on tt.addptr?

bmyerz0 (approved):

test/Conversion/TritonToStructured/gather_scatter_ptr_1d_with_mask.mlir
python3kgae: Added test for if.
Loop not work due to https://github.com/microsoft/triton-shared/pull/332.

bmyerz0 (approved):

PR #325, opened 2025-08-18
Convert Tptr Dialect to LLVM

Realtyxxx:
Directly, I want to solve the problem of tptr dialect drop, mlir-opt can't recognize tptr dialect, so I used pass manager;
Furthermore;
I refer to triton-cpu and triton-linalg to handle the conversion of unstructured ptr types to llvm;
I used **nhat's code case** to test, verified its correctness, and placed it in the test directory

### DIFF
--- a/lib/Conversion/TPtrToLLVM/TPtrToLLVM.cpp
+++ b/lib/Conversion/TPtrToLLVM/TPtrToLLVM.cpp
@@ -5,6 +5,8 @@
 #include "mlir/Dialect/LLVMIR/LLVMTypes.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Ptr/IR/PtrTypes.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/IR/Value.h"
@@ -25,7 +27,6 @@ namespace tptr {
 static bool isOneToOneCast(UnrealizedConversionCastOp op) {
   return (op.getInputs().size() == 1 && op->getNumResults() == 1);
 }
-
 
 // PtrAddOp -> llvm.getelementptr conversion
 struct PtrAddConverter : OpConversionPattern<tptr::PtrAddOp> {
@@ -109,8 +110,10 @@ struct ToMemrefConverter : OpConversionPattern<tptr::ToMemrefOp> {
       }
     }
 
-    Type targetType = getTypeConverter()->convertType(cast<MemRefType>(op.getType()));
-    LDBG("matchAndRewrite: to_memref (typeconverted) " <<  cast<MemRefType>(op.getType()) << " -> " << targetType);
+    Type targetType =
+        getTypeConverter()->convertType(cast<MemRefType>(op.getType()));
+    LDBG("matchAndRewrite: to_memref (typeconverted) "
+         << cast<MemRefType>(op.getType()) << " -> " << targetType);
     if (!targetType) {
       return rewriter.notifyMatchFailure(op, "failed to convert memref type");
     }
@@ -166,14 +169,10 @@ struct FromMemrefConverter : OpConversionPattern<tptr::FromMemrefOp> {
     LDBG("matchAndRewrite: from_memref (before) " << op);
 
     Value input = adaptor.getInput();
+    // 期望此处的输入已通过 TypeConverter 转换为目标 LLVM 结构体类型
     if (isa<MemRefType>(input.getType())) {
-      input = rewriter
-                  .create<UnrealizedConversionCastOp>(
-                      op.getLoc(),
-                      getTypeConverter()->convertType(
-                          cast<MemRefType>(input.getType())),
-                      input)
-                  .getResult(0);
+      return rewriter.notifyMatchFailure(op,
+                                         "expected converted memref descriptor");
     }
 
     // Extract base_ptr (index 0)
@@ -210,7 +209,8 @@ struct UnrealizedCastConverter
     }
 
     if (isa<ptr::PtrType>(outputType) ||
-        (isa<LLVM::LLVMPointerType>(inputType) && isa<MemRefType>(outputType))) {
+        (isa<LLVM::LLVMPointerType>(inputType) &&
+         isa<MemRefType>(outputType))) {
       LDBG("UnrealizedCast (reject): unsafe pointer conversion " << op);
       return rewriter.notifyMatchFailure(op, "unsafe pointer conversion");
     }
@@ -292,9 +292,8 @@ struct ConvertBranchOp : OpConversionPattern<cf::BranchOp> {
       return failure();
     }
 
-    auto newOp =
-        rewriter.replaceOpWithNewOp<cf::BranchOp>(op, op.getDest(),
-                                                  adaptor.getDestOperands());
+    auto newOp = rewriter.replaceOpWithNewOp<cf::BranchOp>(
+        op, op.getDest(), adaptor.getDestOperands());
     LDBG("matchAndRewrite: cf.br (after) -> " << newOp);
     return success();
   }
@@ -341,12 +340,44 @@ struct MemRefAllocConverter : OpConversionPattern<memref::AllocOp> {
       totalElements *= dim;
     }
 
-    // For now, use alloca instead of malloc to avoid complex call setup
-    Value totalSize = rewriter.create<LLVM::ConstantOp>(
+    // Compute total allocation size in bytes = numElements * sizeof(element)
+    Value numElementsVal = rewriter.create<LLVM::ConstantOp>(
         loc, i64Ty, rewriter.getIntegerAttr(i64Ty, totalElements));
 
-    Value allocatedPtr = rewriter.create<LLVM::AllocaOp>(
-        loc, ptrTy, ptrTy, totalSize, /*alignment=*/0);
+    // Query pointer size from DataLayout
+    DataLayout dl = DataLayout::closest(op);
+    auto ptrSize = dl.getTypeSize(ptrTy);
+    if (ptrSize.isScalable()) {
+      return rewriter.notifyMatchFailure(op,
+                                         "scalable pointer size unsupported");
+    }
+    auto fixedPtrSize = static_cast<int64_t>(ptrSize.getFixedValue());
+    Value ptrSizeVal = rewriter.create<LLVM::ConstantOp>(
+        loc, i64Ty, rewriter.getIntegerAttr(i64Ty, fixedPtrSize));
+
+    Value totalBytes =
+        rewriter.create<LLVM::MulOp>(loc, numElementsVal, ptrSizeVal);
+
+    // Declare or lookup malloc: ptr (i64)
+    ModuleOp module = op->getParentOfType<ModuleOp>();
+    auto mallocName = StringRef("malloc");
+    LLVM::LLVMFuncOp mallocFunc =
+        module.lookupSymbol<LLVM::LLVMFuncOp>(mallocName);
+    if (!mallocFunc) {
+      OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToStart(module.getBody());
+      auto mallocType =
+          LLVM::LLVMFunctionType::get(ptrTy, {i64Ty}, /*isVarArg=*/false);
+      mallocFunc =
+          rewriter.create<LLVM::LLVMFuncOp>(loc, mallocName, mallocType);
+    }
+
+    auto mallocCallee = SymbolRefAttr::get(mallocFunc);
+    Value allocatedPtr =
+        rewriter
+            .create<LLVM::CallOp>(loc, TypeRange{ptrTy}, mallocCallee,
+                                  ValueRange{totalBytes})
+            .getResult();
 
     // Build memref descriptor struct
     Value result = rewriter.create<LLVM::UndefOp>(loc, llvmStructType);
@@ -407,10 +438,13 @@ struct MemRefStoreConverter : OpConversionPattern<memref::StoreOp> {
       auto ptrTy = LLVM::LLVMPointerType::get(ctx);
       auto i64Ty = rewriter.getIntegerType(64);
 
-      // Extract base pointer from memref descriptor (index 0)
+      // Extract aligned pointer and offset from memref descriptor
+      // aligned_ptr at index 1, offset at index 2
       Value memrefDescriptor = adaptor.getMemref();
-      Value basePtr = rewriter.create<LLVM::ExtractValueOp>(
-          loc, ptrTy, memrefDescriptor, rewriter.getDenseI64ArrayAttr({0}));
+      Value alignedPtr = rewriter.create<LLVM::ExtractValueOp>(
+          loc, ptrTy, memrefDescriptor, rewriter.getDenseI64ArrayAttr({1}));
+      Value baseOffset = rewriter.create<LLVM::ExtractValueOp>(
+          loc, i64Ty, memrefDescriptor, rewriter.getDenseI64ArrayAttr({2}));
 
       // Calculate linear index from multi-dimensional indices
       Value linearIndex = nullptr;
@@ -425,11 +459,10 @@ struct MemRefStoreConverter : OpConversionPattern<memref::StoreOp> {
                     .getResult(0);
           }
         }
-        linearIndex = index;
+        linearIndex = rewriter.create<LLVM::AddOp>(loc, baseOffset, index);
       } else {
         // Multi-dimensional: linearIndex = i0*stride0 + i1*stride1 + ...
-        linearIndex = rewriter.create<LLVM::ConstantOp>(
-            loc, i64Ty, rewriter.getIntegerAttr(i64Ty, 0));
+        linearIndex = baseOffset;
 
         for (auto [i, index] : llvm::enumerate(adaptor.getIndices())) {
           // Convert index to i64 if needed
@@ -453,8 +486,8 @@ struct MemRefStoreConverter : OpConversionPattern<memref::StoreOp> {
       }
 
       // GEP to get the address of the element
-      Value elementPtr =
-          rewriter.create<LLVM::GEPOp>(loc, ptrTy, ptrTy, basePtr, linearIndex);
+      Value elementPtr = rewriter.create<LLVM::GEPOp>(loc, ptrTy, ptrTy,
+                                                      alignedPtr, linearIndex);
 
       // Store the value
       auto storeOp =
@@ -497,16 +530,24 @@ struct MemRefLoadConverter : OpConversionPattern<memref::LoadOp> {
       auto ptrTy = LLVM::LLVMPointerType::get(ctx);
       auto i64Ty = rewriter.getIntegerType(64);
 
-      // Extract base pointer from memref descriptor (index 0)
+      // Extract aligned pointer and offset from memref descriptor
+      // aligned_ptr at index 1, offset at index 2
       Value memrefDescriptor = adaptor.getMemref();
-      Value basePtr = rewriter.create<LLVM::ExtractValueOp>(
-          loc, ptrTy, memrefDescriptor, rewriter.getDenseI64ArrayAttr({0}));
-
+      LDBG("memrefDescriptor " << memrefDescriptor);
+      Value alignedPtr = rewriter.create<LLVM::ExtractValueOp>(
+          loc, ptrTy, memrefDescriptor, rewriter.getDenseI64ArrayAttr({1}));
+      LDBG("basePtr " << rewriter.create<LLVM::ExtractValueOp>(
+          loc, ptrTy, memrefDescriptor, rewriter.getDenseI64ArrayAttr({0})));
+      LDBG("alignedPtr " << alignedPtr);
+      Value baseOffset = rewriter.create<LLVM::ExtractValueOp>(
+          loc, i64Ty, memrefDescriptor, rewriter.getDenseI64ArrayAttr({2}));
+      LDBG("baseOffset " << baseOffset);
       // Calculate linear index from multi-dimensional indices
       Value linearIndex = nullptr;
       if (adaptor.getIndices().size() == 1) {
         // Single dimension case
         Value index = adaptor.getIndices()[0];
+        LDBG("if index " << index);
         // Convert index to i64 if needed
         if (index.getType() != i64Ty) {
           if (isa<IndexType>(index.getType())) {
@@ -515,11 +556,11 @@ struct MemRefLoadConverter : OpConversionPattern<memref::LoadOp> {
                     .getResult(0);
           }
         }
-        linearIndex = index;
+        linearIndex = rewriter.create<LLVM::AddOp>(loc, baseOffset, index);
       } else {
         // Multi-dimensional: linearIndex = i0*stride0 + i1*stride1 + ...
-        linearIndex = rewriter.create<LLVM::ConstantOp>(
-            loc, i64Ty, rewriter.getIntegerAttr(i64Ty, 0));
+        linearIndex = baseOffset;
+        LDBG("else index " << linearIndex);
 
         for (auto [i, index] : llvm::enumerate(adaptor.getIndices())) {
           // Convert index to i64 if needed
@@ -543,8 +584,8 @@ struct MemRefLoadConverter : OpConversionPattern<memref::LoadOp> {
       }
 
       // GEP to get the address of the element
-      Value elementPtr =
-          rewriter.create<LLVM::GEPOp>(loc, ptrTy, ptrTy, basePtr, linearIndex);
+      Value elementPtr = rewriter.create<LLVM::GEPOp>(loc, ptrTy, ptrTy,
+                                                      alignedPtr, linearIndex);
 
       // Load the value
       Value loadedValue =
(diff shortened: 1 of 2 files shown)


backend/compiler.py
```diff
@@ -78,41 +78,38 @@ def _ttsharedir_to_llir(ttsharedir: str):
...
+        triton_shared.to_llir.add_convert_scf_to_cf(pm)
+        triton_shared.to_llir.add_convert_arith_to_llvm(pm)
+        triton_shared.to_llir.add_convert_math_to_llvm(pm)
+        triton_shared.to_llir.add_convert_complex_to_llvm(pm)
+        triton_shared.to_llir.add_convert_vector_to_llvm(pm)
+        triton_shared.to_llir.add_convert_index_to_llvm(pm)
+        triton_shared.to_llir.add_memref_expand(pm)
+        triton_shared.to_llir.add_finalize_memref_to_llvm(pm)
+        triton_shared.to_llir.add_convert_func_to_llvm(pm)
+        # triton_shared.debug.enable_mlir_debug("tptr-to-llvm")
```
red1bluelost: What is this comment for?

backend/compiler.py
```diff
@@ -78,41 +78,38 @@ def _ttsharedir_to_llir(ttsharedir: str):
         llmlir_path = os.path.join(tmpdir, "ll.mlir")
         llir_path = os.path.join(tmpdir, "ll.ir")
         Path(ttshared_path).write_text(ttsharedir)
-        mlir_opt_path = _get_llvm_bin_path("mlir-opt")
-        # TritonShared-MLIR to LLVM-MLIR
-        subprocess.check_call([mlir_opt_path, ttshared_path,
```
red1bluelost, line 83: Could we preserve all the original comments since they explain why we need certain passes.

lib/Dialect/TPtr/IR/TPtrDialect.cpp
```diff
@@ -39,54 +47,55 @@ void mlir::tptr::TPtrDialect::registerTypes() {
 /// Dialect creation, the instance will be owned by the context. This is the
 /// point of registration of custom types and operations for the dialect.
 void mlir::tptr::TPtrDialect::initialize() {
-  addAttributes<
-#define GET_ATTRDEF_LIST
-#include "triton-shared/Dialect/TPtr/IR/TPtrAttributes.cpp.inc"
-      >();
+  //   addAttributes<
```
red1bluelost: If we are deleting code, please completely remove it and not just comment out.

lib/Conversion/TritonToLinalgExperimental/TritonToPtrPass.cpp
red1bluelost, line 1: This looks like we are allowing Triton IR such as tt.ptr to persist pass any Triton Shared passes be removing the conversions here. We specifically do not want any Triton IR after Triton Shared. This is why we are using tptr as a temporary implementation of the addr/ptr dialect proposal before it is fully upstream in MLIR.

backend/compiler.py
Realtyxxx: This is similar to "--debug-only=tptr-to-llvm", but it can only accept one parameter temporarily. This is what I use when debugging. Sorry, I haven't removed it yet.

backend/compiler.py
Realtyxxx: Of course

Realtyxxx:
> Looking at the general Triton+TritonShared+CPU pipeline. There are three stages:
> 
> 1. Triton Frontend
> 2. Triton Shared middle conversions
> 3. MLIR backend, CPU as an example here but companies likely have a different backend
> 
> We should not add LLVM dialect to stage 1 or 2. This is mainly because it is a very low level dialect and might not mesh as well during the early stages of the stage 3 MLIR backends. Your TptrToLLVM pass would be running somewhere near the end of stage 3 so is fine to generate LLVM dialect.

Ensure we all understand the current plan. Now I am lowering the tptr to llvm at the end of stage 3:

image

So, I wonder if it would be ok for changing tptr to using ptr.ptr type? and not changing the location of pass ?

triton_shared.cc
```diff
@@ -1,8 +1,182 @@
...
+  m.def(name,                                  \
+        [](mlir::PassManager &pm, ty0 val0) { pm.addPass(builder(val0)); })
+
+#define ADD_PASS_WRAPPER_1_ARG(name, builder, ty0, arg0, val0)            \
+  m.def(                                                                  \
+      name,                                                               \
+      [](mlir::PassManager &pm, ty0 arg0) { pm.addPass(builder(val0)); }, \
+      py::arg("pm"), py::arg(#arg0) = val0);
+
+// 一个函数，用于设置 MLIR/LLVM 的调试类型
```
Realtyxxx: to be deleted

triton_shared.cc
```diff
@@ -1,8 +1,182 @@
...
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+
+// MLIR: Target and Translation
+// #include "mlir/Target/LLVMIR/Dialect/AMX/AMXToLLVMIRTranslation.h"
+#include "mlir/Target/LLVMIR/Dialect/Builtin/BuiltinToLLVMIRTranslation.h"
+#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
+
+// llvm:Debug
+#include "llvm/Support/Debug.h"  // 关键头文件
```
Realtyxxx: 注释清理

Realtyxxx:
Hi @red1bluelost I have new changes, Please review again

CMakeLists.txt
```diff
@@ -13,7 +13,41 @@ add_subdirectory(lib)
...
+        MLIRComplexToLLVM
+        MLIRVectorToLLVM
+        MLIRVectorToLLVMPass
+        MLIRFuncToLLVM
+        MLIRControlFlowToLLVM
+        MLIRMemRefToLLVM
+        MLIRVectorToSCF
+        TPtrToLLVM
+        MLIRUBToLLVM
+        # MLIRMathToLibm
```
red1bluelost: Please delete comment

CMakeLists.txt
```diff
@@ -13,7 +13,41 @@ add_subdirectory(lib)
 add_subdirectory(test)
 add_subdirectory(tools/triton-shared-opt)
 
-if (TRITON_SHARED_BUILD_CPU_BACKEND)
+if(TRITON_SHARED_BUILD_CPU_BACKEND)
     add_triton_plugin(TritonShared ${CMAKE_CURRENT_SOURCE_DIR}/triton_shared.cc LINK_LIBS TritonSharedAnalysis TritonToLinalg TritonTilingExtIR)
-    target_link_libraries(TritonShared PRIVATE Python3::Module pybind11::headers ${Python3_LIBRARIES})
+    target_link_libraries(TritonShared
+        PUBLIC
```
red1bluelost: I think that these link libraries should also be private. Please test that out.

include/triton-shared/Conversion/TPtrToLLVM/Passes.td
```diff
@@ -0,0 +1,20 @@
...
+  let options = [
+//       Option<"addptrToLLVM", "ptradd-to-llvm", "bool", /*default*/"true",
+//              "Convert tptr.ptradd on tensors to llvm">,
+//       Option<"fromMemrefToLLVM", "from-memref-to-llvm", "bool", /*default*/"true",
+//              "Convert tptr.from_memref on tensors to llvm">,
+//       Option<"toMemrefToLLVM", "to-memref-to-llvm", "bool", /*default*/"true",
+//              "Convert tptr.to_memref on tensors to llvm">,
+//       Option<"typeoffsetToConst", "typeoffset-to-const", "bool", /*default*/"true",
+//              "Convert tptr.typeoffset to llvm.mlir.const">,
+  ];
```
red1bluelost: Please delete comments then also delete options since it will be empty.

include/triton-shared/Conversion/TPtrToLLVM/Passes.td
```diff
@@ -0,0 +1,20 @@
...
+  let dependentDialects = ["mlir::tptr::TPtrDialect", "mlir::LLVM::LLVMDialect"];
```
red1bluelost: I suspect we will want ptr::PtrDialect included here eventually if not now.

include/triton-shared/Conversion/TPtrToLLVM/TPtrToLLVM.h
```diff
@@ -0,0 +1,32 @@
...
+namespace mlir {
+namespace tptr {
+
+#define GEN_PASS_DECL
+// #define GEN_PASS_CLASSES
+// #define GEN_PASS_DECL_TPTRTOLLVM
+// #define GEN_PASS_DEF_TPTRTOLLVM
+#include "triton-shared/Conversion/TPtrToLLVM/Passes.h.inc"
+
+// void populateTPtrToLLVMCanonicalizationPatterns(RewritePatternSet &patterns);
```
red1bluelost: Delete comments.

include/triton-shared/Dialect/TPtr/IR/TPtrDialect.td
```diff
@@ -116,6 +116,11 @@ def TPTR_TypeOffsetOp : TPTR_Op<"type_offset", [ConstantLike, Pure]> {
      attr-dict $baseType custom<IntType>(type($result))
   }];
   let hasFolder = 1;
+let extraClassDeclaration = [{
```
red1bluelost: Needs leading spaces.

include/triton-shared/Dialect/TPtr/IR/TPtrDialect.td
```diff
@@ -116,6 +116,11 @@ def TPTR_TypeOffsetOp : TPTR_Op<"type_offset", [ConstantLike, Pure]> {
...
+    /// Returns the type offset according to `layout`. If `layout` is `nullopt`
+    /// the nearest layout the op will be used for the computation.
+    llvm::TypeSize getTypeSize(std::optional<DataLayout> layout = std::nullopt);
```
red1bluelost: We should not be adding new functionality to TPtr ops for this. Where this is used make the function local.

lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+      rewriter.replaceOp(op, newCast);
+      return success();
+    }
+    return failure();
+  }
+};
+
+// Type converter for TPtr to LLVM conversion
+class TptrToLLVMTypeConverter : public TypeConverter {
```
red1bluelost: Just use struct and remove 'public:'

lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp
```diff
@@ -0,0 +1,292 @@
...
+public:
+  Type convertPtrPointerType(ptr::PtrType type) {
+    auto ctx = type.getContext();
+    auto pointeeType = type.getElementType();
+
+    if (pointeeType && isa<RankedTensorType>(pointeeType)) {
```
red1bluelost: I do not think we should be seeing tensor types here. I would expect bufferization to have removed them by this point. What scenario are you seeing tensors?

lib/Conversion/TPtrToLLVM/TPtrToLLVM.cpp
```diff
@@ -0,0 +1,614 @@
...
+// PtrAddOp -> llvm.getelementptr conversion
+class PtrAddConverter : public OpConversionPattern<tptr::PtrAddOp> {
+  using OpConversionPattern<tptr::PtrAddOp>::OpConversionPattern;
+
+  Type convertPtrPointerType(ptr::PtrType type,
+                                Type elemTy = nullptr) const {
+    auto ctx = type.getContext();
+    auto pointeeType = elemTy ? elemTy : type.getElementType();
+
+    if (isa<RankedTensorType>(pointeeType)) {
```
red1bluelost: Same comment about seeing tensor types here.

tools/triton-shared-opt/RegisterTritonSharedDialects.h
```diff
@@ -38,14 +39,22 @@ inline void registerTritonSharedDialects(mlir::DialectRegistry &registry) {
...
   registry.insert<
-      mlir::tptr::TPtrDialect, mlir::ptr::PtrDialect,
-      mlir::ttx::TritonTilingExtDialect, mlir::tts::TritonStructuredDialect,
-      mlir::triton::TritonDialect, mlir::cf::ControlFlowDialect,
-      mlir::math::MathDialect, mlir::arith::ArithDialect, mlir::scf::SCFDialect,
-      mlir::gpu::GPUDialect, mlir::linalg::LinalgDialect,
-      mlir::func::FuncDialect, mlir::tensor::TensorDialect,
-      mlir::memref::MemRefDialect, mlir::bufferization::BufferizationDialect>();
+      mlir::tptr::TPtrDialect,
+      // mlir::ptr::PtrDialect,
```
red1bluelost: Please return pointer dialect. We definitely want it here.

include/triton-shared/Conversion/TPtrToLLVM/TPtrToLLVM.h
```diff
@@ -0,0 +1,32 @@
...
+
+void populateTPtrToLLVMConversionPatterns(RewritePatternSet &patterns,
+                                          TypeConverter &typeconverter);
+
+std::unique_ptr<OperationPass<ModuleOp>> createTPtrToLLVMPass();
+
+static bool isOneToOneCast(UnrealizedConversionCastOp op) {
```
red1bluelost: Move this out of the header. It should not be here.

lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp
```diff
@@ -0,0 +1,292 @@
...
+namespace tptr {
+#define GEN_PASS_DEF_TPTRTOLLVM
+#include "triton-shared/Conversion/TPtrToLLVM/Passes.h.inc"
+} // namespace tptr
+} // namespace mlir
+
+namespace {
+
+// Simplify chained UnrealizedConversionCast operations
+struct SimplifyUnrealizedCast
```
red1bluelost: Why do we need to be simplifying this here?

include/triton-shared/Conversion/TPtrToLLVM/Passes.td
Realtyxxx: I previously tried using ptr.ptrtype, and I think when we use ptrdialect, we need to wait until Mr fabianmcg merges most of the ptr-related changes into LLVM.

include/triton-shared/Conversion/TPtrToLLVM/Passes.td
Realtyxxx: It has been fixed, please take another look.

include/triton-shared/Conversion/TPtrToLLVM/TPtrToLLVM.h
Realtyxxx: It has been fixed, please take another look.

include/triton-shared/Dialect/TPtr/IR/TPtrDialect.td
Realtyxxx: This is a function I copied from ptr.typeoffset, I'll move it locally

lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp
Realtyxxx: The first time developing this pass, I noticed similar redundant IR. Just checked and didn't see any similar patterns currently; it will be eliminated

CMakeLists.txt
Realtyxxx: You are correct, private should indeed be used here

Realtyxxx (commented):
already finished all conversation， please check again

CMakeLists.txt
```diff
@@ -14,7 +14,37 @@ add_subdirectory(lib)
...
+        MLIRFuncToLLVM
+        MLIRControlFlowToLLVM
+        MLIRMemRefToLLVM
+        MLIRVectorToSCF
+        TPtrToLLVM
+        MLIRUBToLLVM
+
+        Python3::Module
+        pybind11::headers
+        ${Python3_LIBRARIES}
```
red1bluelost: Please remove Python3_LIBRARIES

backend/compiler.py
red1bluelost: This is not resolved, there are other comments here.

red1bluelost:
Please do not resolve my comments. I will resolve them as I go through a re-review.

include/triton-shared/Conversion/TPtrToLLVM/Passes.td
red1bluelost: Can we add ptr::PtrDialect to the list of 'dependentDialects' here.

lib/Dialect/TPtr/IR/TPtrOps.cpp
```diff
@@ -1,17 +1,18 @@
-#include "mlir/Interfaces/SideEffectInterfaces.h" // Required for IR/TPtrOps.h.inc
 #include "mlir/Bytecode/BytecodeOpInterface.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
```
red1bluelost: Please remove all edits from this file.

include/triton-shared/Dialect/TPtr/IR/TPtrDialect.td
```diff
@@ -215,3 +215,4 @@ def TTPTR_StoreOp : TPTR_Op<"store", [
 }
 
 #endif // TPTR_DIALECT
+
```
red1bluelost: Please remove edits from this file.

include/triton-shared/Conversion/TPtrToLLVM/TPtrToLLVM.h
```diff
@@ -0,0 +1,22 @@
...
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "triton-shared/Dialect/TPtr/IR/TPtrDialect.h"
+
+namespace mlir {
+namespace tptr {
+
+#define GEN_PASS_DECL
+#include "triton-shared/Conversion/TPtrToLLVM/Passes.h.inc"
+void populateTPtrToLLVMConversionPatterns(RewritePatternSet &patterns,
```
red1bluelost, line 15: (nit) Put a new line between the include and function declaration.

lib/Conversion/TPtrToLLVM/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+  TPtrTableGen
+
+  LINK_LIBS PUBLIC
+  MLIRIR
+  MLIRPass
+  MLIRTransforms
+  MLIRSupport
+  MLIRReconcileUnrealizedCasts
+  TPtrIR
+  MLIRDialectUtils
```
red1bluelost: (nit) alphabetical order

tools/triton-shared-opt/RegisterTritonSharedDialects.h
```diff
@@ -38,14 +39,22 @@ inline void registerTritonSharedDialects(mlir::DialectRegistry &registry) {
...
-      mlir::func::FuncDialect, mlir::tensor::TensorDialect,
-      mlir::memref::MemRefDialect, mlir::bufferization::BufferizationDialect>();
+      mlir::tptr::TPtrDialect,
+      mlir::ptr::PtrDialect,
+      mlir::ttx::TritonTilingExtDialect,
+      mlir::tts::TritonStructuredDialect,
+      mlir::triton::TritonDialect,
+      mlir::cf::ControlFlowDialect, mlir::scf::SCFDialect,
+      mlir::math::MathDialect, mlir::arith::ArithDialect,
+      // mlir::gpu::GPUDialect,
```
red1bluelost: Delete this.

red1bluelost: Also please run clang-format

red1bluelost (changes requested):
Please rebase your changes onto an updated Triton Shared.

Realtyxxx:
Sorry, I will start making changes based on your suggestion now



> Please do not resolve my comments. I will resolve them as I go through a re-review.

include/triton-shared/Conversion/TPtrToLLVM/Passes.td
Realtyxxx: Previously I thought we only used ptrType so we didn't add PtrDialect, but of course we could

backend/compiler.py
Realtyxxx: Sorry I didn't understand your meaning well before, I will add it back

lib/Dialect/TPtr/IR/TPtrOps.cpp
Realtyxxx: I have replaced it back to the original file style. Just to confirm, do I need to rebase back to the original commit and delete all modifications to the file in the commit to prevent the existence of commit records?

include/triton-shared/Dialect/TPtr/IR/TPtrDialect.td
Realtyxxx: Same question as the previous conversation

Dasor:
> @Dasor
> 
> 1. `memref.alloc` should be converted to `malloc`, which is the way it should, because the pass that was converted to memref before did not construct free, so there was no error; This has been modified, by the way, @red1bluelost  Is there any plan for this memory leak?
> 2. The memoryspace in PtrType may be introduced in the future, using the conversion method you recommended, and minor modifications will also require conversion when there is no memSpace.
> 3. Can you give more specific advice about `MemRefLoadConverter` and `MemRefStoreConverter`.


1. Yes, there will be a memory leak if there is no dealloc pass (currently there is not such pass in the CPU pipeline of triton shared) but there should be (just add the --buffer-deallocation-pipeline pass) and in that case it produces a crash.
2. You can try and do the conversion from the Tptr memoryspace instead.
3. If you implement the type conversion you don't not need `MemRefLoadConverter` and `MemRefStoreConverter` (not even `MemRefAllocConverter`) as LLVM will take care of it automatically using the `TypeConverter` when it sees a memref of type `ptr`

Realtyxxx:
The branch has been rebased and squashed. Please take another look. @red1bluelost @Dasor

Dasor:
The malloc looks good, however, there is no conversion for `memref.dealloc` still I think all the memrefs conversion patterns can be avoided by the TypeConverter but I see yours seems to not be doing anything (as it should be converting dealloc) I think it's because rather than using your own TypeConverter you need to use LLVMTypeConverter from `lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp` and add the conversion there but I'm not 100% sure.

you can use this example to test and see if `tptr-to-llvm` is working correctly:

```mlir
func.func @kernel(%ptr: !ptr.ptr<#ptr.generic_space>) {
      %c0 = arith.constant 0 : index
      %c1 = arith.constant 1 : index
      %c128 = arith.constant 128 : index
      %alloc_4 = memref.alloc() {alignment = 64 : i64} : memref<128x!ptr.ptr<#ptr.generic_space>>
      scf.for %arg9 = %c0 to %c128 step %c1 {
        memref.store %ptr, %alloc_4[%arg9] : memref<128x!ptr.ptr<#ptr.generic_space>>
      }
      memref.dealloc %alloc_4 : memref<128x!ptr.ptr<#ptr.generic_space>>
      return
    }
}
```

Realtyxxx:
> The malloc looks good, however, there is no conversion for `memref.dealloc` still I think all the memrefs conversion patterns can be avoided by the TypeConverter but I see yours seems to not be doing anything (as it should be converting dealloc) I think it's because rather than using your own TypeConverter you need to use LLVMTypeConverter from `lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp` and add the conversion there but I'm not 100% sure.malloc 看起来不错，但是仍然没有 `memref.dealloc` 的转换。我认为所有的 memrefs 转换模式都可以通过 TypeConverter 避免，但我看到你的似乎什么都没做（它应该转换 dealloc）。我想是因为你不需要使用自己的 TypeConverter，而应该使用来自 `lib/Conversion/TPtrToLLVM/TPtrToLLVMPass.cpp` 的 LLVMTypeConverter，并在那里添加转换，但我不是 100%确定。
> 
> you can use this example to test and see if `tptr-to-llvm` is working correctly:你可以使用这个示例来测试并查看 `tptr-to-llvm` 是否正常工作：
> 
> ```mlir
> func.func @kernel(%ptr: !ptr.ptr<#ptr.generic_space>) {
>       %c0 = arith.constant 0 : index
>       %c1 = arith.constant 1 : index
>       %c128 = arith.constant 128 : index
>       %alloc_4 = memref.alloc() {alignment = 64 : i64} : memref<128x!ptr.ptr<#ptr.generic_space>>
>       scf.for %arg9 = %c0 to %c128 step %c1 {
>         memref.store %ptr, %alloc_4[%arg9] : memref<128x!ptr.ptr<#ptr.generic_space>>
>       }
>       memref.dealloc %alloc_4 : memref<128x!ptr.ptr<#ptr.generic_space>>
>       return
>     }
> }
> ```

using LLVMTypeConverter didn't change anything, ptr-to-llvm is handled along with other transformations in the convert-to-llvm pass, where it doesn't handle memref situations all at once

Realtyxxx:
> Well I think that's the point if LLVM already handles that with `convert-to-llvm` there is no need to redo the work. We just need to add that pass to the pass pipeline and let `tptr-to-llvm` handle only the `tptr` dialect without changing anything related to the `ptr` dialect.
> 
> I think once #331 gets merged and #329, #327 gets done the only part of TPtr we are going to need is `IntToPtr` and `PtrToInt`

I have a similar opinion. In the future, the tptr dialect in triton-shared may be replaced by the improved ptr dialect. It seems that my changes will be irrelevant.

PR #332, opened 2025-08-21
Add example for row-gather's unsupported scenario

nhat-nguyen:


bmyerz0:
referencing the related issue #334

PR #333, opened 2025-08-22
Add support for unstructured masks in MaskAnalysis

python3kgae:
- Enhanced `MaskAnalysis` to handle unstructured masks. MaskState will save unstructured mask for a dimension if that dimension failed MaskAnalysis.
- Updated `TritonStructuredDialect` to include `gather_scatter_mask` in `MakeGatherScatterTensorPtrOp`.
- Modified `PtrAnalysis` to apply unstructured masks during pointer analysis.
- Extended `StructuredToMemref` conversion to support generic masks in load/store operations.
- Added verification logic for `MakeGatherScatterTensorPtrOp` to ensure compatibility with masks.

### DIFF
--- a/lib/AnalysisStructured/PtrAnalysis.cpp
+++ b/lib/AnalysisStructured/PtrAnalysis.cpp
@@ -53,22 +53,23 @@ static Value applyUnstructuredMask(Operation *op, Value ptr,
   }
 
   auto [dim, unstructuredMask] = masks[0];
-  if (auto scatterPtr =
+  if (auto gatherScatterPtr =
           ptr.getDefiningOp<tts::MakeGatherScatterTensorPtrOp>()) {
-    if (dim != scatterPtr.getGatherScatterDim()) {
+    if (dim != gatherScatterPtr.getGatherScatterDim()) {
       op->emitRemark("MaskAnalysis failed for unstructured mask dim not equal "
                      "gather scatter dim");
       return nullptr;
     }
 
-    ptr = builder
-              .create<tts::MakeGatherScatterTensorPtrOp>(
-                  loc, scatterPtr.getBase(),
-                  scatterPtr.getGatherScatterOffset(), unstructuredMask,
-                  scatterPtr.getGatherScatterDim(), scatterPtr.getSizes(),
-                  scatterPtr.getMixedStrides(), scatterPtr.getMixedOffsets())
-              .getResult();
-
+    ptr =
+        builder
+            .create<tts::MakeGatherScatterTensorPtrOp>(
+                loc, gatherScatterPtr.getBase(),
+                gatherScatterPtr.getGatherScatterOffset(), unstructuredMask,
+                gatherScatterPtr.getGatherScatterDim(),
+                gatherScatterPtr.getSizes(), gatherScatterPtr.getMixedStrides(),
+                gatherScatterPtr.getMixedOffsets())
+            .getResult();
   } else if (auto structuredPtr = ptr.getDefiningOp<tts::MakeTensorPtrOp>()) {
     auto ofrToI32Value = [&](OpFoldResult ofr) {
       Value v = dyn_cast<Value>(ofr);
--- a/test/Conversion/TritonToStructured/unstructured_mask_2d_non_continuous_store_kernel.mlir
+++ b/test/Conversion/TritonToStructured/unstructured_mask_2d_non_continuous_store_kernel.mlir
@@ -1,6 +1,7 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
-// Make sure make_gather_scatter_tptr with unsturctured mask generate correctly from column-structured ptr with unstructured mask.
+// Make sure make_gather_scatter_tptr with unstructuredmask generate correctly from column-structured ptr with unstructured mask.
+// The store is unstructured ptr on dim 1 and unstructured mask on dim 1.
 
 // CHECK-LABEL:   tt.func public @generic_mask_2d_non_continuous_store_kernel(
 // CHECK-SAME:                                                                %[[VAL_0:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !tt.ptr<f32> {tt.divisibility = 16 : i32},
--- a/test/Conversion/TritonToStructured/unstructured_mask_3d_non_continuous_load_kernel.mlir
+++ b/test/Conversion/TritonToStructured/unstructured_mask_3d_non_continuous_load_kernel.mlir
@@ -1,6 +1,7 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
 // Make sure make_gather_scatter_tptr with generic mask generate correctly from row-structured ptr with unstructured mask.
+// The load is unstructured ptr on dim 1 and unstructured mask on dim 1.
 
 // CHECK-LABEL:   tt.func public @generic_mask_3d_non_continuous_load_kernel(
 // CHECK-SAME:                                                                %[[VAL_0:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !tt.ptr<f32> {tt.divisibility = 16 : i32},
--- a/test/Conversion/TritonToStructured/unstructured_mask_2d_kernel.mlir
+++ b/test/Conversion/TritonToStructured/unstructured_mask_2d_kernel.mlir
@@ -1,6 +1,8 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
-// Make sure make_gather_scatter_tptr with unsturctured mask generate correctly from structured ptr with unstructured mask.
+// Make sure make_gather_scatter_tptr with unstructuredmask generate correctly from structured ptr with unstructured mask.
+// The load is structured ptr, with unstructured mask on dim 0.
+// The store is structured ptr, with unstructured mask on dim 1.
 
 // CHECK-LABEL:   tt.func public @generic_mask_2d_kernel(
 // CHECK-SAME:                                           %[[VAL_0:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !tt.ptr<f32> {tt.divisibility = 16 : i32},
--- a/test/Conversion/TritonToStructured/unstructured_mask_2d_non_continuous_load_kernel.mlir
+++ b/test/Conversion/TritonToStructured/unstructured_mask_2d_non_continuous_load_kernel.mlir
@@ -1,6 +1,7 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
-// Make sure make_gather_scatter_tptr with unsturctured mask generate correctly from row-structured ptr with unstructured mask.
+// Make sure make_gather_scatter_tptr with unstructuredmask generate correctly from row-structured ptr with unstructured mask.
+// The load is unstructured ptr on dim 0 and unstructured mask on dim 0.
 
 // CHECK-LABEL:   tt.func public @generic_mask_2d_non_continuous_load_kernel(
 // CHECK-SAME:                                                               %[[VAL_0:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !tt.ptr<f32> {tt.divisibility = 16 : i32},
--- a/test/Conversion/TritonToStructured/unstructured_mask_3d_kernel.mlir
+++ b/test/Conversion/TritonToStructured/unstructured_mask_3d_kernel.mlir
@@ -1,6 +1,8 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
 // Make sure make_gather_scatter_tptr with generic mask generate correctly from structured ptr with unstructured mask.
+// The load is structured ptr, with unstructured mask on dim 1.
+// The store is structured ptr, with unstructured mask on dim 2.
 
 // CHECK-LABEL:   tt.func public @generic_mask_3d_kernel(
 // CHECK-SAME:                                           %[[VAL_0:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !tt.ptr<f32> {tt.divisibility = 16 : i32},
--- a/test/Conversion/TritonToStructured/unstructured_mask_3d_non_continuous_store_kernel.mlir
+++ b/test/Conversion/TritonToStructured/unstructured_mask_3d_non_continuous_store_kernel.mlir
@@ -1,6 +1,7 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
 // Make sure make_gather_scatter_tptr with generic mask generate correctly from column-structured ptr with unstructured mask.
+// The store is unstructured ptr on dim 2 and unstructured mask on dim 2.
 
 
 // CHECK-LABEL:   tt.func public @generic_mask_3d_non_continuous_store_kernel(


python/examples/test_generic_mask.py
```diff
@@ -0,0 +1,284 @@
+import torch
```
bmyerz0: Is "generic" used in this way throughout the codebase prior to this PR? If not, then I wonder if we can introduce a more descriptive name than "generic mask".

Other ideas include: irregular mask, unstructured mask, unliftable mask, fallback mask. Not all of these are good ideas :) just throwing out a bunch

python/examples/test_generic_mask.py
```diff
@@ -0,0 +1,284 @@
...
+
+import triton
+import triton.language as tl
+
+from triton.backends.triton_shared.driver import CPUDriver
+
+@triton.jit
+def generic_mask_2d_kernel(in_ptr, out_ptr, mask_m_ptr, mask_n_ptr, m, n, M:tl.constexpr, N:tl.constexpr):
```
bmyerz0: can you add comments for which dimensions are expected to be contiguous or not

include/triton-shared/Analysis/MaskAnalysis.h
```diff
@@ -44,17 +44,33 @@ namespace triton {
...
+// Bool tensor mask could be saved into masks in case that dimension failed
+// MaskAnalysis. These is to allow case where only one dimension failed while
+// others passed. A MakeGatherScatterTensorPtrOp operation could be generated
+// for the failed dimension. Only 3 patterns are supported for this.
+// 1. offsets[:, None] < n where the offsets is 1d tensor.
+//    It will in pattern of expandDims -> broadcast -> cmp
+// 2. mask[:, None] where mask is 1d bool tensor.
+//    It will in pattern of cmp -> expandDims -> broadcast
+// 3. scalar_mask[:, None] where scalar mask is scalar bool.
+//    It will in pattern of splat -> expandDims -> broadcast
```
bmyerz0, line 57: If I understand, these are cases where one dimension can use the lifted mask and another dimension cannot?

Is this why we only support 3 patterns instead of arbitrary tensor of bools?

test/Conversion/TritonToStructured/generic_mask_2d_kernel.mlir
```diff
@@ -0,0 +1,95 @@
+// RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
+
+// Make sure make_gather_scatter_tptr with generic mask generate correctly.
```
bmyerz0: these comments are the same for all 2d tests. Can you differentiate them so it is clear at a glance what the difference is?

test/Conversion/TritonToStructured/generic_mask_2d_non_continuous_load_kernel.mlir
```diff
@@ -0,0 +1,84 @@
...
+// CHECK:           %[[VAL_27:.*]] = arith.minsi %[[VAL_26]], %[[VAL_8]] : index
+// CHECK:           %[[VAL_28:.*]] = tts.make_gather_scatter_tptr %[[VAL_0]] to sizes: [4, 16] gather_scatter_dim: 0 gather_scatter_offset: %[[VAL_19]] gather_scatter_mask: %[[VAL_22]], strides: {{\[}}%[[VAL_23]], 1], offsets: [0, 0] : tensor<4xi32> tensor<4xi1> <f32> to !tt.ptr<tensor<4x16xf32>>
+// CHECK:           %[[VAL_29:.*]] = "tts.load"(%[[VAL_28]], %[[VAL_27]], %[[VAL_7]]) <{operandSegmentSizes = array<i32: 1, 1, 1>, static_mask_dims = array<i64: 0, -9223372036854775808>}> : (!tt.ptr<tensor<4x16xf32>>, index, f32) -> tensor<4x16xf32>
+// CHECK:           %[[VAL_30:.*]] = tts.make_tptr %[[VAL_1]] to sizes: [4, 16], strides: {{\[}}%[[VAL_23]], 1], offsets: [0, 0], shape: [0, 0], order: [] : <f32> to tensor<4x16x!tt.ptr<f32>>
+// CHECK:           %[[VAL_31:.*]] = arith.minsi %[[VAL_18]], %[[VAL_11]] : index
+// CHECK:           "tts.store"(%[[VAL_30]], %[[VAL_29]], %[[VAL_31]], %[[VAL_27]]) <{static_mask_dims = array<i64: -9223372036854775808, -9223372036854775808>}> : (tensor<4x16x!tt.ptr<f32>>, tensor<4x16xf32>, index, index) -> ()
```
bmyerz0: is it necessary to check the end of this kernel after the tts.load val28?

test/Conversion/TritonToStructured/generic_mask_2d_kernel.mlir
```diff
@@ -0,0 +1,95 @@
...
+// CHECK:           %[[VAL_24:.*]] = "tts.load"(%[[VAL_20]], %[[VAL_23]], %[[VAL_8]]) <{operandSegmentSizes = array<i32: 1, 1, 1>, static_mask_dims = array<i64: -9223372036854775808>}> : (tensor<16x!tt.ptr<i8>>, index, i8) -> tensor<16xi8>
+// CHECK:           %[[VAL_25:.*]] = arith.extsi %[[VAL_24]] : tensor<16xi8> to tensor<16xi32>
+// CHECK:           %[[VAL_26:.*]] = arith.cmpi ne, %[[VAL_25]], %[[VAL_11]] : tensor<16xi32>
+// CHECK:           %[[VAL_27:.*]] = arith.minsi %[[VAL_23]], %[[VAL_7]] : index
+// CHECK:           %[[VAL_28:.*]] = tt.make_range {end = 8 : i32, start = 0 : i32} : tensor<8xi32>
+// CHECK:           %[[VAL_29:.*]] = tts.make_gather_scatter_tptr %[[VAL_0]] to sizes: [8, 16] gather_scatter_dim: 0 gather_scatter_offset: %[[VAL_28]] gather_scatter_mask: %[[VAL_19]], strides: {{\[}}%[[VAL_7]], 1], offsets: [0, 0] : tensor<8xi32> tensor<8xi1> <f32> to !tt.ptr<tensor<8x16xf32>>
+// CHECK:           %[[VAL_30:.*]] = "tts.load"(%[[VAL_29]], %[[VAL_27]], %[[VAL_6]]) <{operandSegmentSizes = array<i32: 1, 1, 1>, static_mask_dims = array<i64: 0, -9223372036854775808>}> : (!tt.ptr<tensor<8x16xf32>>, index, f32) -> tensor<8x16xf32>
```
bmyerz0: what is the difference between this one and *non_continuous_load.mlir?

lib/Analysis/MaskAnalysis.cpp
```diff
@@ -361,7 +403,45 @@ LogicalResult MaskState::parseExtSI(arith::ExtSIOp op, const Location loc,
 LogicalResult MaskState::parseCmp(arith::CmpIOp cmpOp, const Location loc,
                                   OpBuilder &builder) {
   assert(this->isEmpty());
-
+  int cmpOpDim = -1;
+  if (auto shapedType = dyn_cast<ShapedType>(cmpOp.getType())) {
+    for (unsigned r = 0; r < shapedType.getRank(); r++) {
+      if (shapedType.getShape()[r] != 1) {
+        if (cmpOpDim != -1) {
+          cmpOpDim = -1;
```
bmyerz0, line 459: is this a case where there are two such dims? can you comment it

lib/Analysis/MaskAnalysis.cpp
```diff
@@ -361,7 +403,45 @@ LogicalResult MaskState::parseExtSI(arith::ExtSIOp op, const Location loc,
...
+      if (shapedType.getRank() > 1) {
+        // If cmpOp is not 1D, collapse it to 1D.
+        auto flatType = RankedTensorType::get({shapedType.getShape()[cmpOpDim]},
+                                              shapedType.getElementType());
+        auto maybeReassociationMap =
+            getReassociationIndicesForReshape(shapedType, flatType);
+        SmallVector<ReassociationIndices> reassociation =
+            *maybeReassociationMap;
+        // Set masks.
+        genericMask = builder.create<tensor::CollapseShapeOp>(
```
bmyerz0: why is this needed?

lib/Analysis/MaskAnalysis.cpp
```diff
@@ -449,7 +529,10 @@ LogicalResult MaskState::parseCmp(arith::CmpIOp cmpOp, const Location loc,
     else
       this->dims.push_back(lhsState.dims[i]);
   }
-
+  if (cmpOpDim != -1) {
+    // Clear masks when success.
+    masks[cmpOpDim] = nullptr;
```
bmyerz0, line 582: why is this needed?

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
```diff
@@ -885,6 +885,11 @@ struct LoadConverter : public OpConversionPattern<tts::LoadOp> {
       if (auto gatherMaskDimIndex = getIntAttr(gatherMaskDim)) {
         // If the gather mask dimension is a constant, we can use it directly.
         unsigned gatherMaskDimValue = gatherMaskDimIndex.value();
+        if (gatherMaskDimValue == 0 && ptr.getGatherScatterMask()) {
+          // For generic mask case, do the full loop and use the generic mask to
```
bmyerz0: what does "do the full loop" mean

python/examples/test_generic_mask.py
```diff
@@ -0,0 +1,284 @@
...
+    offs_m = tl.arange(0, M)
+    offs_n = tl.arange(0, N)
+
+    mask_m = tl.load(mask_m_ptr + offs_m, mask=offs_m < m, other=0) != 0
+    mask_n = tl.load(mask_n_ptr + offs_n, mask=offs_n < n, other=0) != 0
+
+    in_ptrs = in_ptr + offs_m[:, None] * N + offs_n[None, :]
+    v = tl.load(in_ptrs, mask=mask_m[:, None] and offs_n[None, :] < n, other=-2)
+    out_ptrs = out_ptr + offs_m[:, None] * N + offs_n[None, :]
+    tl.store(out_ptrs, v, mask=offs_m[:, None] < m and mask_n[None, :])
```
bmyerz0: are the load and store two cases of mixed scalar and generic mask? Maybe the test name and/or comments should say this

python/examples/test_generic_mask.py
```diff
@@ -0,0 +1,284 @@
...
+    M = triton.next_power_of_2(m)
+    N = triton.next_power_of_2(n)
+    input = torch.arange(2, 2 + (m * n), device=device, dtype=torch.float32).reshape(m, n)
+    output = torch.full_like(input, -1)
+    mask_m = torch.tensor([1, 0, 1, 0, 1, 0], device=device, dtype=torch.bool)
+    
+    mask_n = torch.tensor([0, 1, 0, 1, 0, 1, 0, 1,
+                            1, 0, 1, 0, 1, 0, 1, 0], device=device, dtype=torch.bool)
+
+    generic_mask_2d_kernel[1, 1, 1](input, output, mask_m.to(torch.int8), mask_n.to(torch.int8), m, n, M, N)
```
bmyerz0: nit: I think [(1,)] is more idiomatic in typical programs

python/examples/test_generic_mask.py
```diff
@@ -0,0 +1,284 @@
...
+    index_n = torch.tensor([10, 20, 15, 1, 3, 5, 7, 0, 2, 4, 6, 8], device=device, dtype=torch.int32)
+
+    stride_b = input.stride(0)
+    stride_m = input.stride(1)
+    stride_n = input.stride(2)
+
+    I_N = triton.next_power_of_2(len(index_n))
+
+    output = torch.full((b, m, n), -1, device=device, dtype=torch.float32)
+    generic_mask_3d_non_continuous_store_kernel[1, 1, 1](input, output, index_n, len(index_n), I_N, M, B, b, m, n, stride_b, stride_m, stride_n)
```
bmyerz0: it seems it would be helpful for at least one test where the kernel's address and kernel's mask calculation depend on program id, since that is what we will actually see in the wild.

python/examples/test_generic_mask.py
```diff
@@ -0,0 +1,284 @@
...
+    torch.testing.assert_close(output, expected_output)
+
+
+@triton.jit
+def generic_mask_3d_non_continuous_store_kernel(in_ptr, out_ptr, index_n_ptr, i_n, I_N:tl.constexpr, M:tl.constexpr, B:tl.constexpr, b, m, n, stride_b, stride_m, stride_n):
+    offs_m = tl.arange(0, M)
+    offs_n = tl.arange(0, I_N)
+    offs_b = tl.arange(0, B)
+
+    index_n = tl.load(index_n_ptr + offs_n, mask=offs_n < n, other=0)
```
bmyerz0: at a glance it seems all tests compute the mask with tl.load. Are there other tensor ops to try as well? For example just doing some arbitrary math on tensors of 1's, or 0's or range?

python/examples/test_generic_mask.py
bmyerz0: IIUC, all these tests should have worked already by relying on the pointer fallback path right? 
Or maybe they didn't because there is no CPU backend lowering for pointer fallback path?
Even so, how can we keep these tests future-proof if CPU backend support for pointer fallback is added?

Can we also check that the gather scatter path is actually being taken?

lib/Analysis/MaskAnalysis.cpp
```diff
@@ -628,18 +719,75 @@ LogicalResult MaskState::parseExpandDims(triton::ExpandDimsOp expandDimsOp,
...
   this->dims.insert(this->dims.begin() + axis, builder.getIndexAttr(1));
 
   return success();
 }
 
+// Return all non-nullptr masks along with their dimensions.
+SmallVector<std::pair<unsigned, Value>> MaskState::getGenericMasks() {
+  SmallVector<std::pair<unsigned, Value>> result;
+
+  for (auto [i, m] : llvm::enumerate(masks)) {
```
bmyerz0, line 832: nit: can this be done with a filter?

python3kgae: Failed for llvm::enumerate cannot cast to std::pair.

python/examples/test_generic_mask.py
python3kgae: Yes. All tests should work by the fallback path.

Added check for the gather scatter generated.

python/examples/test_generic_mask.py
python3kgae: Added one cmp with program id

python/examples/test_generic_mask.py
python3kgae: Added.

python/examples/test_generic_mask.py
python3kgae: Updated.

python/examples/test_generic_mask.py
python3kgae: Added comments.

lib/Conversion/StructuredToMemref/StructuredToMemref.cpp
python3kgae: Updated to `loop on every elements`

lib/Analysis/MaskAnalysis.cpp
python3kgae: We want to know the difference between the mask passed the analysis or not.
If it passed, we'll save nullptr. If it failed, we'll save the unstructured mask.

lib/Analysis/MaskAnalysis.cpp
python3kgae: This happens when the cmp on shape like [1x2] which will have mask [1x2].
But we only want to save 1D mask.
So we collapse it to 1D here.

lib/Analysis/MaskAnalysis.cpp
python3kgae: Added comment.
```
          // This will happen when the cmp has more than one dimension with size
          // larger than 1.
          // Like a < b while both a and b are tensors with shape 2x2.
```

test/Conversion/TritonToStructured/generic_mask_2d_kernel.mlir
python3kgae: non_continuous_load have non continuous ptr with unstructured mask.
this one have continuous ptr with unstructured mask.

test/Conversion/TritonToStructured/generic_mask_2d_non_continuous_load_kernel.mlir
python3kgae: No.
Removed check after the load.

test/Conversion/TritonToStructured/generic_mask_2d_kernel.mlir
python3kgae: Updated.

include/triton-shared/Analysis/MaskAnalysis.h
python3kgae: These patterns are about how a bool tensor generated from 1D or scalar bool.
The bool value itself could be from arbitrary expression.

python/examples/test_generic_mask.py
python3kgae: Done.

python/examples/test_generic_mask.py
python3kgae: Switched to unstructured mask which matches ptr analysis.

lib/AnalysisStructured/PtrAnalysis.cpp
```diff
@@ -37,6 +37,91 @@
 
 #define DEBUG_TYPE "triton-ptr-analysis"
 
+using namespace mlir;
+
+// Try to apply unstructured mask on the ptr.
+static Value applyUnstructuredMask(Operation *op, Value ptr,
```
bmyerz0, line 43: thanks for renaming the tptr. Can you also rename the scatterPtr as we discussed?

test/Conversion/TritonToStructured/unstructured_mask_2d_non_continuous_store_kernel.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
-// Make sure make_gather_scatter_tptr with generic mask generate correctly.
+// Make sure make_gather_scatter_tptr with unsturctured mask generate correctly from column-structured ptr with unstructured mask.
```
bmyerz0: nit typo unstructured

test/Conversion/TritonToStructured/unstructured_mask_3d_non_continuous_load_kernel.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: triton-shared-opt --triton-to-structured --remove-dead-values --cse --canonicalize %s | FileCheck %s
 
-// Make sure make_gather_scatter_tptr with generic mask generate correctly.
+// Make sure make_gather_scatter_tptr with generic mask generate correctly from row-structured ptr with unstructured mask.
```
bmyerz0, line 3: Can you make this more precise? "row-structured ptr with unstructured mask." is ambiguous.
- is the tensor's pointer arithmetic structured or unstructured for each of the row and col dims?
- is the unstructured mask being applied to the row or col dim?

same comment for other lit test

python3kgae: Done.

test/Conversion/TritonToStructured/unstructured_mask_2d_non_continuous_store_kernel.mlir
python3kgae: Fixed.

lib/AnalysisStructured/PtrAnalysis.cpp
python3kgae: Fixed.

bmyerz0 (approved):

PR #344, opened 2025-08-26
Remove triton submodule and add triton-hash.txt to track the triton commit ID

lechenyu:
This PR removes the Triton submodule and replaces it with a hash file (`triton-hash.txt`) to track the Triton version. This change provides greater flexibility in the Triton versions that triton-shared can be built with. All relevant files have been updated accordingly.

After this PR is landed, the installation steps will be 
```sh
export TRITON_PLUGIN_DIRS=$(pwd)/triton_shared

git clone https://github.com/microsoft/triton-shared.git triton_shared
git clone https://github.com/triton-lang/triton.git
cd triton && git checkout $(cat ../triton_shared/triton-hash.txt)
...
// follow the same steps as before to build Triton
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,6 @@
 compile_commands.json
 build/*
 .vscode/*
-.clang-format
+/.clang-format
 test_core.py
 test_annotations.py


triton-san/script/install_triton_san.sh
```diff
@@ -3,17 +3,17 @@
 #!/bin/bash
 set -e
 
-if [ "$#" -lt 3 ]; then
-  echo "Usage: $0 <desired path for triton-san installation> <path to venv> <path to LLVM install dir>"
+if [ "$#" -lt 4 ]; then
+  echo "Usage: $0 <desired path for triton-san installation> <path to LLVM install directory> <path to Python venv> <path to triton>"
```
red1bluelost: Could you specify that this is path to Triton source code/repository.

.github/workflows/test-plugin.yml
```diff
@@ -39,12 +39,18 @@ jobs:
       with:
         ref: ${{ inputs.triton-shared-ref }}
         path: triton_shared
-        # Also checkout triton submodule
-        submodules: recursive
+
+    - name: Checkout triton based on triton-hash.txt
```
red1bluelost, line 43: Is it possible to do this checkout with an action. I'm not sure if it is best practice to manually close. But I do not know actions well.

red1bluelost:
Seems I cannot add non-Microsoft reviewers so just tagging folks in this comment. @haishanzzzz, @mitekoth, @zhaoshiz

red1bluelost:
@lechenyu Right now we have three symlinks into Triton: 'python/examples/test_core.py', 'python/examples/test_annotations.py', and '.clang-format'. For the clang-format, I'd be fine with having a copy in our repo. I doubt it would change in a way we care about. For the two test files, we need a way to import them into examples so they can be tested.

lechenyu:
> @lechenyu Right now we have three symlinks into Triton: 'python/examples/test_core.py', 'python/examples/test_annotations.py', and '.clang-format'. For the clang-format, I'd be fine with having a copy in our repo. I doubt it would change in a way we care about. For the two test files, we need a way to import them into examples so they can be tested.

Apologies for missing those symlinks. I was thinking it might be helpful to set up a top-level CMakeLists.txt to automate the build process. Since we're trying to remove the Triton submodule, there are now several additional steps that users need to be aware of, such as checking out Triton manually and making the two Python test visible to triton-shared.

lechenyu:
I’ve updated the CMakeLists.txt to automatically generate the three symlinks. These symlinks are now included in .gitignore.

CMakeLists.txt
```diff
@@ -18,3 +18,16 @@ if (TRITON_SHARED_BUILD_CPU_BACKEND)
...
 endif()
+
+# Add symlinks to selected pytest files and the clang-format setting in triton. The tests are imported into triton-shared’s test folder to 
+# run under triton-shared's conftest configuration, and the clang-format link ensures consistent code style enforcement across both repositories.
+cmake_path(APPEND CMAKE_CURRENT_SOURCE_DIR "python" "examples" "test_core.py" OUTPUT_VARIABLE TRITON_SHARED_TEST_CORE)
+cmake_path(APPEND CMAKE_CURRENT_SOURCE_DIR "python" "examples" "test_annotations.py" OUTPUT_VARIABLE TRITON_SHARED_TEST_ANNOTATIONS)
+cmake_path(APPEND CMAKE_CURRENT_SOURCE_DIR ".clang-format" OUTPUT_VARIABLE TRITON_SHARED_CLANG_FORMAT_SETTING)
+cmake_path(APPEND CMAKE_SOURCE_DIR "python" "test" "unit" "language" "test_core.py" OUTPUT_VARIABLE TRITON_TEST_CORE)
+cmake_path(APPEND CMAKE_SOURCE_DIR "python" "test" "unit" "language" "test_annotations.py" OUTPUT_VARIABLE TRITON_TEST_ANNOTATIONS)
+cmake_path(APPEND CMAKE_SOURCE_DIR ".clang-format" OUTPUT_VARIABLE TRITON_CLANG_FORMAT_SETTING)
```
red1bluelost, line 37: This will likely only work if the base project is Triton. I think we should check if the original file exists at that location then do the symlink, otherwise do not symlink. I think that would be the best way. If the base project is not Triton then you would not get symlinks which is fine.

CMakeLists.txt
```diff
@@ -18,3 +18,16 @@ if (TRITON_SHARED_BUILD_CPU_BACKEND)
...
+
+file(CREATE_LINK ${TRITON_TEST_CORE} ${TRITON_SHARED_TEST_CORE} SYMBOLIC)
+file(CREATE_LINK ${TRITON_TEST_ANNOTATIONS} ${TRITON_SHARED_TEST_ANNOTATIONS} SYMBOLIC)
+file(CREATE_LINK ${TRITON_CLANG_FORMAT_SETTING} ${TRITON_SHARED_CLANG_FORMAT_SETTING} SYMBOLIC)
```
red1bluelost: I think for clang-format it is preferable to just keep around a copy rather than try to symlink it.

.gitignore
```diff
@@ -2,4 +2,7 @@
 .cache
 compile_commands.json
 build/*
-.vscode/*
\ No newline at end of file
+.vscode/*
+.clang-format
```
red1bluelost: Remove clang-format from here since we are going to make it a hard copy.

red1bluelost (changes requested):

CMakeLists.txt
lechenyu: I think using symlink is correct here, since the purpose is to make sure triton-shared uses the same format as triton.

CMakeLists.txt
lechenyu: Fixed.

.gitignore
lechenyu: Same as before. Hard copy is not necessary.

triton-san/script/install_triton_san.sh
lechenyu: Fixed.

.gitignore
red1bluelost: If the hard copy is not necessary, we should update this so that it only matched the one in the base directory and not any in subdirectories.

.github/workflows/test-plugin.yml
lechenyu: I did some research using the GitHub documentation as well as ChatGPT. It looks like we can use actions to check out the Triton repo, but switching to a specific commit ID (stored in the `triton-hash.txt` file) requires additional steps in the CI/CD setup (see the code block below).
Personally, I’m not a fan of either approach below since it still involves running bash commands or using GitHub scripts, which feels like overkill for the issue we’re trying to solve. That said, I’m okay with moving forward with these fixes if it’s the preferred direction.

```
jobs:
  checkout-from-file:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout current repo
        uses: actions/checkout@v4

      - name: Read commit hash from file
        id: commit_id
        run: |
          commit=$(cat triton-hash.txt)   # your file path here
          echo "commit_id=$commit" >> "$GITHUB_OUTPUT"

      - name: Checkout target repo at that commit
        uses: actions/checkout@v4
        with:
          repository: triton-lang/triton
          ref: ${{ steps.commit_id.outputs.commit_id }}
          path: triton
          fetch-depth: 0   # optional: get full history
```
or
```
jobs:
  self-checkout-at-file-commit:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/github-script@v7
        id: read_hash
        with:
          script: |
            const path = 'commit-hash.txt'; // <-- your file path
            const {owner, repo} = context.repo;
            const res = await github.rest.repos.getContent({ owner, repo, path, ref: context.sha });
            const commit = Buffer.from(res.data.content, res.data.encoding).toString('utf8').trim();
            core.setOutput('commit', commit);

      - uses: actions/checkout@v4
        with:
          ref: ${{ steps.read_hash.outputs.commit }}
          fetch-depth: 0

```

.gitignore
lechenyu: Just updated.

.github/workflows/test-plugin.yml
red1bluelost: I'm fine with leaving it the way it is. I am fine with the github script idea, but we can leave that for a future improvement of the actions if we need to refactor them at some point.

red1bluelost (approved):

PR #345, opened 2025-08-27
Update triton to e44bd1c83c1c3e8deac7c4f02683cfb3cc395c8b

red1bluelost:
This updates to latest Triton main which should resolve the nightly build issues.

Primary changes:
* Change conversion of tts::MakeGatherScatterTensorPtrOp to insert identity cast so that DialectConversion does not ignore the type converted operand of the associated tts::LoadOp when supplying the op adaptor to the pattern.
* Move support of unsplat to the new dedicated operation since they do not do reduction anymore
* Add missing test figure to conftest
* Disable float annotation tests since bfloat16 and float16 are not supported in CPU backend
* Add missing link library to registering of passes
* Update construction of ValueRange{} which was causing compilation errors.
* Remove unused builder in TPtrOps since it caused linker errors.
* Remove bfloat16 and float16 from CPU backend since not testing is present. Better to crash during compilation than get runtime errors.
* Remove GPUDialect that is no longer present upstream

### DIFF
--- a/backend/driver.py
+++ b/backend/driver.py
@@ -54,6 +54,10 @@ def _ty_to_cpp(ty):
         "u16": "uint16_t",
         "u32": "uint32_t",
         "u64": "uint64_t",
+        # Proper support for bfloat16 and float16 is not yet handled.
+        # https://github.com/microsoft/triton-shared/issues/348
+        # "fp16": "TODO",
+        # "bf16": "TODO",
         "fp32": "float",
         "f32": "float",
         "fp64": "double",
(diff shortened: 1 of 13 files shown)


backend/driver.py
```diff
@@ -54,8 +54,6 @@ def _ty_to_cpp(ty):
         "u16": "uint16_t",
         "u32": "uint32_t",
         "u64": "uint64_t",
-        "fp16": "float",
```
lechenyu, line 57: Shall we just comment them and leave a Todo here? I am not sure whether we have any plan to fix it in the future.

lechenyu (commented):
This PR looks good to me overall. Just a few minor comments.

backend/driver.py
lechenyu: Has discussed this offline and will tackle it in the future.

lechenyu (approved):

PR #346, opened 2025-08-28
Fix MaskAnalysis bool type workaround

lxwangruc:
In MaskAnalysis, a Boolean type shouldn't be passed around as Index.
This replaces the workaround added in PR https://github.com/microsoft/triton-shared/pull/318. 
We use the following two tests as baseline, they needed the workaround to pass before and now passing with our fix.
```
test/Conversion/TritonToStructured/mask_ld_st_scalar_dim.mlir
python/examples/test_mask.py
```

### DIFF
--- a/lib/Analysis/MaskAnalysis.cpp
+++ b/lib/Analysis/MaskAnalysis.cpp
@@ -302,9 +302,13 @@ LogicalResult MaskState::parseConstant(arith::ConstantOp constOp,
 LogicalResult MaskState::parseIntScalar(Value scalar, const Location loc,
                                         OpBuilder &builder) {
   assert(this->isEmpty());
-  auto castOp =
-      builder.create<arith::IndexCastOp>(loc, builder.getIndexType(), scalar);
-  this->scalar = castOp.getResult();
+  if (scalar.getType().isInteger(1)) {
+    this->scalar = scalar;
+  } else {
+    auto castOp =
+        builder.create<arith::IndexCastOp>(loc, builder.getIndexType(), scalar);
+    this->scalar = castOp.getResult();
+  }
   return success();
 }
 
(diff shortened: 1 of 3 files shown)


lib/Analysis/MaskAnalysis.cpp
```diff
@@ -302,9 +302,13 @@ LogicalResult MaskState::parseConstant(arith::ConstantOp constOp,
 LogicalResult MaskState::parseIntScalar(Value scalar, const Location loc,
                                         OpBuilder &builder) {
   assert(this->isEmpty());
-  auto castOp =
-      builder.create<arith::IndexCastOp>(loc, builder.getIndexType(), scalar);
-  this->scalar = castOp.getResult();
+  if (!(scalar.getType().isInteger(1))) {
```
red1bluelost: Please flip this condition and remove the extra parenthesis.

lib/Analysis/OpFoldResultUtils.cpp
```diff
@@ -52,6 +52,18 @@ bool hasConstZero(const OpFoldResult ofr) {
   return false;
 }
 
+Value ofrToValue(const OpFoldResult ofr, const Location loc, OpBuilder &b) {
+  if (Value val = dyn_cast<Value>(ofr)) {
+    return val;
+  }
+  auto intVal = getIntAttr(ofr);
+  if (intVal.has_value()) {
+    return b.create<arith::ConstantOp>(loc, b.getIndexAttr(intVal.value()));
```
red1bluelost: This should not be index attr since it should be using the type from ofr.

red1bluelost: Instead of having this if statement, we could just return a constant op where we take in the ofr casted to Attribute then casted to TypedAttr.

red1bluelost (approved):

PR #347, opened 2025-08-29
[TritonArithToLinalg] Preserve function visibility during conversion.

red1bluelost:
For cases where functions are not inlined we want to preserve visibility information.


beicy (approved):
LGTM!

PR #349, opened 2025-09-02
Setup clang-format action and format the whole codebase.

red1bluelost:
This PR does the following:

* Clang-Formats all C and C++ files within our repository using the LLVM style.
* Creates a github workflow that will verify code formatting on PRs.
* Removes symlink of clang-format in favor of a permanent copy. We do not care to keep it up to date with Triton. The likelihood of Triton updating their clang format file is low enough for this to not be a problem. Even if it were updated, it is not essential for us to fully match their style. It is more important to have a permanent copy that keeps the workflows simple.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,27 @@
+name: clang-format
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install clang-format==20.1.8 ripgrep
+    - name: Running clang-format
+      run: |
+        rg . --type cpp --type c --files-with-matches \
+            | xargs clang-format --dry-run --Werror
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -8,19 +8,15 @@ on:
 jobs:
   clang-format:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.12"]
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python 3.12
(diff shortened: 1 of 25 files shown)


lxwangruc (approved):

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install clang-format==20.1.8 ripgrep
```
NathanielMcVicar: Probably safer to pin ripgrep version as well.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,27 @@
...
+        python-version: ["3.12"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
```
NathanielMcVicar: Is there a specific reason we need to upgrade pip here? Otherwise relying on the base version from setup python is probably more reliable.

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
```
NathanielMcVicar: Why use a matrix with a single version?

NathanielMcVicar (changes requested):
This is excellent! I have some very nitty comments on the github action.

.github/workflows/clang-format.yml
red1bluelost: Removed. Mostly just a biproduct of the example I adapted.

.github/workflows/clang-format.yml
red1bluelost: Removed the matrix. Mostly just a biproduct of the example I adapted.

.github/workflows/clang-format.yml
red1bluelost: Pinned the version and checked it locally.

NathanielMcVicar (approved):
LGTM! Thanks

PR #352, opened 2025-09-09
Add support for tt.dot using 3d tensors

Dasor:
Really simple change to emit `linalg.batch_matmul` in the input to `tt.dot` is a 3d tensor.


Dasor:
Hi @ziliangzl yes there are, I added support for it because this test from triton-cpu was failing:

https://github.com/triton-lang/triton-cpu/blob/main/python/test/unit/language/test_core.py#L4203

ziliangzl:
Oh I see, I honestly didn’t know that a 3D `tt.dot` could be implemented this way.  
Thanks a lot for pointing me to the test, that’s really helpful!🙏

PR #354, opened 2025-09-12
Add GatherOp support

ziliangzl:
Added support for `triton::GatherOp` conversion. Using cf.assert for out-of-bound indices, consistent with the CUDA backend behavior (which triggers a device assert).


test/Conversion/TritonToLinalg/gather.mlir
```diff
@@ -0,0 +1,31 @@
+// RUN: triton-shared-opt --triton-to-linalg %s | FileCheck %s
+module {
```
bmyerz0, line 2: are there CHECKs you can add to be sure the translation is as expected?

test/Conversion/TritonToLinalg/gather.mlir
```diff
@@ -0,0 +1,31 @@
...
+    %16 = tt.splat %arg1 : !tt.ptr<i64> -> tensor<8x4x!tt.ptr<i64>>
+    %17 = tt.addptr %16, %15 : tensor<8x4x!tt.ptr<i64>>, tensor<8x4xi32>
+    %18 = tt.load %17 : tensor<8x4x!tt.ptr<i64>>
+    %19 = tt.gather %9[%18] {axis = 0 : i32} : (tensor<4x4xf32>, tensor<8x4xi64>) -> tensor<8x4xf32>
+    %20 = tt.splat %arg2 : !tt.ptr<f32> -> tensor<8x4x!tt.ptr<f32>>
+    %21 = tt.addptr %20, %15 : tensor<8x4x!tt.ptr<f32>>, tensor<8x4xi32>
+    tt.store %21, %19 : tensor<8x4x!tt.ptr<f32>>
+    tt.return
+  }
+}
```
bmyerz0, line 31: Since this PR lowers gather to linalg, that means it should work on the triton-shared CPU backend. I suggest adding a pytest to ensure functional correctness of the lowering

bmyerz0: see https://github.com/microsoft/triton-shared/tree/main/python/examples

test/Conversion/TritonToLinalg/gather.mlir
ziliangzl: Sorry, I indeed forgot to add the CHECKs. They’ve been added now.

test/Conversion/TritonToLinalg/gather.mlir
ziliangzl: I copied the test from test_core.py and added it as test_gather.py. Not sure if this is OK.

ziliangzl:
@microsoft-github-policy-service agree

enjustli:
why not convert `triton::GatherOp` to `tts::GatherOp` 🤔

ziliangzl:
Hi @bmyerz0, just wanted to check if my understanding is correct: the current lowering of triton::GatherOp to linalg is fine, and the only concern is the naming of tts::GatherOp to avoid confusion. Please let me know if there’s anything else I should address. Thanks!

bmyerz0:
> Hi @bmyerz0, just wanted to check if my understanding is correct: the current lowering of triton::GatherOp to linalg is fine, and the only concern is the naming of tts::GatherOp to avoid confusion. Please let me know if there’s anything else I should address. Thanks!

Yes, my main issue was I do not think it is good to lower triton::GatherOp to tts::GatherOp. The suggestion is to lower triton::GatherOp directly to linalg.

bmyerz0:
Yes, for the sake of backends, I think it is reasonable to make any tts op lowering to linalg.generic optional. But we should still provide the lowering. And I think that tts should include two different ops, one from tl.load vs tl.gather.

python/examples/test_gather.py
```diff
@@ -0,0 +1,66 @@
...
+    out_offs = tl.arange(0, out_dim0)
+    tl.store(out_ptr + out_offs, out)
+
+
+@pytest.mark.interpreter
+@pytest.mark.parametrize("src_shape, indices_shape, axis", [
+    ([32], [64], 0),
+    ([4, 4], [8, 4], 0),
+    ([128, 64], [256, 64], 0),
+    ([128, 64], [128, 128], 1),
```
bmyerz0, line 45: these all appear increase the size of the tensor. Can you have test cases that contract the size of the tensor?

PR #359, opened 2025-10-25
Fix TritonSan build error

lechenyu:
This PR resolves the TritonSan build error. The build script was broken due to several issues:

- Removal of the Triton submodule: We now use triton-hash.txt to record the Triton version. The previous logic for locating Triton submodule was invalidated by this change.

- LLVM-related issue: In the corresponding LLVM version, there is a known bug that causes [a compilation error when building OpenMP](https://github.com/llvm/llvm-project/commit/62ff9ac4c68f48c089528105259c68943ab176de). The build script now cherry-picks this patch to address the problem.

This PR fixes all these issues. The build script will now automatically check out Triton and place it alongside triton-shared, LLVM, and the Python virtual environment. There are no changes to the usage of the build script or the TritonSan driver script.

### DIFF
--- a/triton-san/script/build_llvm.sh
+++ b/triton-san/script/build_llvm.sh
@@ -49,6 +49,12 @@ LLVM_HASH=$(cat "${LLVM_HASH_FILE}")
 cd "${LLVM_SOURCE_DIR}"
 git checkout ${LLVM_HASH}
 
+# Cherry-pick the patch to resolve the OpenMP build error. An incorrect setting in openmp/runtime/src/CMakeLists.txt 
+# causes the generated omp-tools.h to be placed in Clang's include directory instead of OpenMP's, leading to a 
+# compilation failure when locating omp-tools.h.
+# More details are available at this link: https://github.com/llvm/llvm-project/commit/62ff9ac4c68f48c089528105259c68943ab176de
+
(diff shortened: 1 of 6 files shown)


triton-san/script/build_llvm.sh
```diff
@@ -49,6 +49,11 @@ LLVM_HASH=$(cat "${LLVM_HASH_FILE}")
 cd "${LLVM_SOURCE_DIR}"
 git checkout ${LLVM_HASH}
 
+if ! git merge-base --is-ancestor 62ff9ac HEAD; then
```
bmyerz0, line 58: 1) comment summary for what the fix is to give context
2) issue-todo to remove this once llvm hash is known to be the next llvm->triton merge beyond this commit. At that point we cannot continue to support old LLVM_HASH

bmyerz0 (changes requested):

triton-san/script/build_llvm.sh
lechenyu: Just add more description and a TODO.

On a related note, the latest Triton already uses a newer LLVM version beyond this commit. Should we consider updating the triton-shared-associated Triton version soon (i.e., test triton-shared with a newer Triton release and update triton-hash.txt.)?

bmyerz0 (approved):

PR #363, opened 2025-11-15
Bump to triton-lang/triton@dbfbc1e1e6c

enjustli:
This updates to latest Triton main which should resolve the nightly build issues.

Primary changes:
* Change llvm/mlir api usage
* Update driver.py


enjustli:
this PR needs https://github.com/triton-lang/triton/pull/8766 merged

enjustli:
@nhat-nguyen, sorry for the explicit ping but I did not have access to request a formal review. Could you be pleaase review the PR / or request review from appropriate stakeholders?
Thanks!

enjustli:
gently pin

nhat-nguyen:
the api changes for op creation and driver.py look good to me. thanks a lot for helping us update this! as for the lit tests, do you know what are the main changes that caused the differences?

tangentially, we're not using the triton-to-linalg pass anymore (triton-to-linalg-experimental is the main one now), so it might be cleaner to remove the pass and those tests first before the triton bump so that we don't have to update the lit files.

enjustli:
> the api changes for op creation and driver.py look good to me. thanks a lot for helping us update this! as for the lit tests, do you know what are the main changes that caused the differences?
> 
> tangentially, we're not using the triton-to-linalg pass anymore (triton-to-linalg-experimental is the main one now), so it might be cleaner to remove the pass and those tests first before the triton bump so that we don't have to update the lit files.

In the new llvm version, `memref.reinterpret_cast` enhances the analysis of `static offset`. so the output `IR` changed. due to too many changed mlir files, so i use `llvm-project/mlir/utils/generate-test-checks.py` to generate  mlir `CHECK` directly.

removing the triton-to-linalg pass helps a lot! 🎉

enjustli:
already update.

enjustli:
gently pin

enjustli:
ping @bmyerz0 @red1bluelost

nhat-nguyen:
i think it would be cleaner if we delete the TritonToLinalg pass in a different PR

enjustli:
> i think it would be cleaner if we delete the TritonToLinalg pass in a different PR

I see. 👌

enjustli:
this pr depends on https://github.com/microsoft/triton-shared/pull/366

bmyerz0:
Please see https://github.com/microsoft/triton-shared/pull/367

